PR #5, opened 2015-01-21
Updated for 1.1.0.Beta2, as 'examples' wasn't working out of the box

jpkrohling:


jorgemoralespou:
Hi jpkrohling, mgoldmann,
I've seen your conversation regarding making the images smaller and in sync, and I have a suggestion, that we are implementing in apiman

Download the artifacts (war files) either from maven central, if published, or from any public location they can be downloaded from. If they are not available in maven central, there could be possibly be available in the keycloak git repository, so you could get them as raw.

I hope it helps, as I'm also struggling with these problems with these images that we are also using for apiman.

jpkrohling:
@jorgemoralespou , we are actually doing that for the main artifacts, namely, the Keycloak adapter and war-dist, but the problem is that the examples need to be built by maven, as they are not distributed in packaged form. Currently, we are building the examples locally and putting the war files in the git repository, using then the FILE directive on the Dockerfile. That helps keep the image with just the required size, but we may get stale example files in the future again.

goldmann:
Closing, since I'm going to merge #6.

PR #10, opened 2015-04-02
port publish mentioned on readme

matzew:
adding -p option to access the server

### DIFF
--- a/server-mysql/README.md
+++ b/server-mysql/README.md
@@ -14,7 +14,7 @@ First start a MySQL instance using the MySQL docker image:
 
 Start a Keycloak instance and connect to the MySQL instance:
 
-    docker run --name keycloak --link mysql:mysql jboss/keycloak-mysql
+    docker run -it -p 8080:8080 --name keycloak --link mysql:mysql jboss/keycloak-mysql
 
 ### Environment variables
 


server-mysql/README.md
```diff
@@ -14,7 +14,7 @@ First start a MySQL instance using the MySQL docker image:
 
 Start a Keycloak instance and connect to the MySQL instance:
 
-    docker run --name keycloak --link mysql:mysql jboss/keycloak-mysql
+    docker run -it -p 8080:8080 --name keycloak --link mysql:mysql jboss/keycloak-mysql
```
goldmann: I would better not commit this, but maybe @mstruk or @abstractj can say what they prefer?

goldmann: Or @stianst, of course :)

stianst:
We should make it consistent for all images, but it makes more sense to me to drop '-p 8080:8080' from the main Keycloak image than to add it here. The readme for the WildFly image doesn't have it and we should be consistent. I reckon that's what makes sense as binding to localhost:8080 prevents running multiple WF/KC instances.

stianst:
Decided to remove -it -p 8080:8080 from main image as well, so closing this

PR #24, opened 2015-07-11
Fix POSTGRES_USER bug that doesn't fit the API in README.md

ayr-ton:
The POSTGRES_USER variable changed to POSTGRES_USERNAME, and this is breaking the API in README.md.

This is causing breaks in automated updates in production environments, that expects to use POSTGRES_USER, the default variable from postgres container.


goldmann:
I'll rebuild the image tomorrow when I get back from PTO.

ayr-ton:
Thank you, @goldmann :muscle:

Issue #25, opened 2015-07-13
Saxon not importing database settings at 1.3.1.Final


ayr-ton:
At version 1.3.1, at the second step of the Dockerfile:

RUN java -jar /usr/share/java/saxon.jar -s:/opt/jboss/keycloak/standalone/configuration/standalone.xml -xsl:/opt/jboss/keycloak/changeDatabase.xsl -o:/opt/jboss/keycloak/standalone/configuration/standalone.xml; java -jar /usr/share/java/saxon.jar -s:/opt/jboss/keycloak/standalone/configuration/standalone-ha.xml -xsl:/opt/jboss/keycloak/changeDatabase.xsl -o:/opt/jboss/keycloak/standalone/configuration/standalone-ha.xml; rm /opt/jboss/keycloak/changeDatabase.xsl

saxon is not writing the changeDatabase.xsl data into /opt/jboss/keycloak/standalone/configuration/standalone.xml or standalone-ha.xml.

No error output is given even if I manually run this step inside the container. It act like everything is fine but when I open the standalone.xml, this is the default settings.

I could not figure out yet a solution for this in meaning to provide a PR. Does someone know's how to fix this?

vst:
I just bumped into this, too. Please check this out: #26. Very simple to make it work back again.

ayr-ton:
@vst Thank you very much! 

![](https://images.duckduckgo.com/iu/?u=http%3A%2F%2Fwww.trcpodcast.com%2Fwp-content%2Fuploads%2F2013%2F03%2Fdawson-crying.jpg)

PR #26, opened 2015-07-14
Fix to datasources namespace (v2.0 -> v3.0)

vst:
It seems that the stylesheet to update the standalon(-ha).xml references to datasources namespace with version 2.0 but it actually is 3.0 at least with Keycloak v1.3.1.Final.


vst:
Must fix Issue #25

PR #27, opened 2015-08-19
Exposing needed port in documentation

mariane-sm:
It happened to me and other coworkers more than once: run the image and forget that since port 8080 is not exposed, it gets impossible to access it.

### DIFF
--- a/server/README.md
+++ b/server/README.md
@@ -6,7 +6,7 @@ Example Dockerfile with Keycloak server.
 
 To boot in standalone mode
 
-    docker run jboss/keycloak
+    docker run -p 8080:8080 jboss/keycloak
 
 Once it boots, you can login to the admin console using admin/admin for the first login.
 


stianst:
Rejecting this - you can still access the image using the IP associated with the image (see docker inspect to retrieve that). If you expose the port on localhost you can only run one image associated with 8080 so it's not a good default. Further, the documentation should only cover details specific to the image, not general information about Docker.

mariane-sm:
Thanks for the quick feedback! So what about letting people know this is the way to go? Most official images states in the readme how to use the image even though it is general information about docker, in order to make even people without much knowledge in docker able to run it as quickly as possible. 

See for example: https://hub.docker.com/_/tomcat/, https://hub.docker.com/_/node/, https://hub.docker.com/_/jenkins/, https://hub.docker.com/_/jetty/.

stianst:
Some official images do, others don't. Finding ip address of an image and exposing ports locally is basic knowledge that you'll learn in any getting started with Docker guides. Again, we're not documenting Docker here, we're only documenting Keycloak specific parts.

PR #36, opened 2015-10-30
add docker-compose.yml

techgaun:
Added a docker-compose based orchestration example.

### DIFF
--- a/server-mongo/README.md
+++ b/server-mongo/README.md
@@ -26,3 +26,6 @@ When starting the Keycloak instance you can pass a number of environment variabl
 
 Specify name of MongoDB database (optional, default is `keycloak`).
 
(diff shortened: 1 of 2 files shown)


stianst:
Thanks for the PR, but I'm going to reject it. Reasoning behind this is that Docker Compose is out of the scope of these Docker files and the readme files. There are also several alternative technologies to Docker Compose out there. In general we don't want to cover to much in the images and readme files as they are mainly a starting point to using Keycloak with Docker. We may however at some point add more detailed documentation on how to use Keycloak with Docker in our documentation. That would be a perfect place to put an example on how to use Docker compose.

techgaun:
No problem, thanks for the details.

PR #41, opened 2016-01-12
Updates for KC 1.8

stianst:


xied75:
:+1: So when can we start to use this from Docker Hub? (I thought KC 1.8 is not released yet?)

stianst:
Images on Docker Hub will be rebuilt once 1.8 is released, which is probably going to be tomorrow

PR #43, opened 2016-01-30
updating from Wildfly 9.0.1 to 10.0.0

ahus1:
updating wildfly version at least to 9.0.2 (didn't test WF 10 yet)


ahus1:
Updated pull request for WF 10.0.0

PR #44, opened 2016-03-02
KEYCLOAK-2573 - Improve Docker Examples image

cmoulliard:
The existing https://github.com/jboss-dockerfiles/keycloak/blob/1.9.0.Final/examples/Dockerfile Dockerfile is not longer up to date to run Keycloak 1.9 + examples
Some improvements are required :

* Move root image to jdk8
* Use maven version > 3.1.1
* Change the location of the keycloak-demo archive as sourceforge doesn't host them
* Pass additional parameter to configure wildfly (""-bmanagement", "0.0.0.0"")
*Expose port 9990
* Add an admin user & keycloak user to access respectively the wildfly console / keycloak admin console


examples/Dockerfile
```diff
@@ -1,18 +1,29 @@
...
 
-RUN cd /opt/jboss && curl -L http://sourceforge.net/projects/keycloak/files/$KEYCLOAK_VERSION/keycloak-demo-$KEYCLOAK_VERSION.tar.gz/download | tar zx && mv /opt/jboss/keycloak-demo-$KEYCLOAK_VERSION /opt/jboss/keycloak-demo
+RUN cd /opt/jboss && curl -s http://downloads.jboss.org/keycloak/$KEYCLOAK_VERSION/keycloak-demo-$KEYCLOAK_VERSION.zip -o tmp.zip && unzip tmp.zip -d . && mv /opt/jboss/keycloak-demo-$KEYCLOAK_VERSION /opt/jboss/keycloak-demo
 
 RUN mvn package -f /opt/jboss/keycloak-demo/examples/preconfigured-demo/pom.xml && rm -rf ~/.m2/repository
 RUN cd /opt/jboss/keycloak-demo/examples/preconfigured-demo && find -name *.war | grep -v ear | xargs -I {} cp {} /opt/jboss/keycloak-demo/keycloak/standalone/deployments/ && cp /opt/jboss/keycloak-demo/examples/preconfigured-demo/testrealm.json /opt/jboss/keycloak-demo/keycloak/
 
-EXPOSE 8080
-CMD ["/opt/jboss/keycloak-demo/keycloak/bin/standalone.sh", "-b", "0.0.0.0", "-Dkeycloak.import=/opt/jboss/keycloak-demo/keycloak/testrealm.json"]
+EXPOSE 8080 9990
```
stianst: Don't expose 9990 by default

examples/Dockerfile
```diff
@@ -1,18 +1,29 @@
...
+
+RUN /opt/jboss/keycloak-demo/keycloak/bin/add-user.sh -u admin -p admin --silent
+RUN /opt/jboss/keycloak-demo/keycloak/bin/add-user-keycloak.sh -u admin -p admin
```
stianst: Don't hard-code users. We've recently removed hard-coded users for a reason.

examples/Dockerfile
```diff
@@ -1,18 +1,29 @@
...
 ENV KEYCLOAK_VERSION 1.9.0.Final
 ENV JBOSS_HOME /opt/jboss/keycloak-demo/keycloak
+ENV MAVEN_VERSION 3.3.9
 
 USER root
 
-RUN yum -y install maven && yum clean all
+RUN yum -y install wget && yum clean all
+
+RUN curl -fsSL https://archive.apache.org/dist/maven/maven-3/$MAVEN_VERSION/binaries/apache-maven-$MAVEN_VERSION-bin.tar.gz | tar xzf - -C /usr/share \
```
stianst: Why are you installing Maven manually rather than just through yum as it was before?

examples/README.md
```diff
@@ -12,6 +12,8 @@ Domain mode is not supported on this version, as this image ships with a custom
 
 Once it boots, you can login using admin/admin for the first login on the [Auth server](http://localhost:8080/auth/admin/) or as bburke@redhat.com/password on the [Customer Portal sample application](http://localhost:8080/customer-portal/customers/view.jsp) 
 
+You can also log on to the [Widlfly console](http://localhost:9990/console/) using admin/admin as login/password.
```
stianst: WildFly console shouldn't be exposed by default

examples/Dockerfile
cmoulliard: There is a rule within the pom which check if the version of maven used is > 3.1.1. As yum installs the rpm 3.0.5 which is older, it is required to install maven as proposed

examples/Dockerfile
cmoulliard: Why ! As the Keycloak demo is deployed top of wildfly, this is important to have also access to the JavaEE server console as you can if you run the project on your machine without docker

examples/Dockerfile
cmoulliard: I will include env variables to pass the KEYCLOAK_USER/KEYCLOAK_PASSWORD to be used

examples/README.md
cmoulliard: See my previous comment

examples/Dockerfile
stianst: Take a look at https://github.com/jboss-dockerfiles/wildfly/blob/master/Dockerfile. If the WildFly docker image doesn't expose 9990 we shouldn't either.

examples/Dockerfile
stianst: OK. We don't really need mvn 3.1.1+ for examples, but as it extends keycloak-parent (which it shouldn't) it ends up requiring. I also figured yum would have a newer version of mvn by now. So we can keep this as you've done it at least for now.

examples/Dockerfile
goldmann: None of the management consoles should be exposed by default. You cannot "unexpose" a port in Docker. Since this is a general use image - only strictly required ports should be exposed. If someone wants to connect to console - it should be exposed by hand.

examples/Dockerfile
```diff
@@ -1,18 +1,29 @@
...
+FROM jboss/base-jdk:8
 
 ENV KEYCLOAK_VERSION 1.9.0.Final
 ENV JBOSS_HOME /opt/jboss/keycloak-demo/keycloak
+ENV MAVEN_VERSION 3.3.9
 
 USER root
 
-RUN yum -y install maven && yum clean all
+RUN yum -y install wget && yum clean all
```
goldmann: This is unused.

examples/Dockerfile
```diff
@@ -1,18 +1,34 @@
...
 
 RUN mvn package -f /opt/jboss/keycloak-demo/examples/preconfigured-demo/pom.xml && rm -rf ~/.m2/repository
 RUN cd /opt/jboss/keycloak-demo/examples/preconfigured-demo && find -name *.war | grep -v ear | xargs -I {} cp {} /opt/jboss/keycloak-demo/keycloak/standalone/deployments/ && cp /opt/jboss/keycloak-demo/examples/preconfigured-demo/testrealm.json /opt/jboss/keycloak-demo/keycloak/
 
+ADD docker-entrypoint.sh /opt/jboss/
+
 EXPOSE 8080
-CMD ["/opt/jboss/keycloak-demo/keycloak/bin/standalone.sh", "-b", "0.0.0.0", "-Dkeycloak.import=/opt/jboss/keycloak-demo/keycloak/testrealm.json"]
+
+USER root
```
stianst: You shouldn't need to do chmod on docker-entrypoint.sh. See https://github.com/jboss-dockerfiles/keycloak/blob/master/server/Dockerfile#L16.

examples/docker-entrypoint.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+
+if [ $KEYCLOAK_USER ] && [ $KEYCLOAK_PASSWORD ]; then
+    /opt/jboss/keycloak-demo/keycloak/bin/add-user.sh -u $KEYCLOAK_USER -p $KEYCLOAK_PASSWORD --silent
```
stianst: This is adding a WildFly console user, not Keycloak admin user. Those are two different things. So please remove this line and only keep the add-user-keycloak

examples/README.md
```diff
@@ -7,10 +7,18 @@ This image prepares a basic Keycloak Auth Server and deploys the examples.
...
-
-Domain mode is not supported on this version, as this image ships with a custom standalone.xml file. 
-
-Once it boots, you can login using admin/admin for the first login on the [Auth server](http://localhost:8080/auth/admin/) or as bburke@redhat.com/password on the [Customer Portal sample application](http://localhost:8080/customer-portal/customers/view.jsp) 
+    
+The user and password to be used to access the Keycloak Admin console can be passed as env variable when you request the creation of the docker process
+       
+    docker run -e KEYCLOAK_USER=admin -e KEYCLOAK_PASSWORD=admin jboss/keycloak-examples
+       
+The docker process can also be started in detached mode and associated to a name
```
stianst: Please don't add generic Docker instructions to the readme. There is plenty of instructions on how to use Docker elsewhere, so we don't need to duplicate this.

cmoulliard: I have just added the instructions to add the KEYCLOAK_USER & KEYCLOAK_PASSWORD

stianst: You've also added instructions on how to "started in detached mode and associated to a name"

cmoulliard: As you don't like that, I will remove it

examples/docker-entrypoint.sh
cmoulliard: No problem

examples/Dockerfile
cmoulliard: This has been required otherwise a permission denied was reported

stianst: Do chmod+x before you commit to github and it should be fine. It works just fine for the server Dockerfile.

cmoulliard: Last modification pushed. That should be perfect now.

stianst:
Great, thanks

PR #45, opened 2016-03-10
mysql ssl environment options

gambol99:
- adding the environment variables MYSQL_SSL and MYSQL_SSL_VERIFY as a means to enforce SSL and verification requirement of cert


stianst:
Thanks for the PR, but I'm going to reject it. We want to keep the base images simple without to many configuration options. There's two main reasons for this. Firstly, we would have to maintain, document and test all these different configuration options, which we don't have the resources to do. Secondly, the images are base images and are intended for you to extend, not to use directly. In most situations you need to extend the image in either case to add a truststore.

PR #46, opened 2016-04-05
Use docker-compose instead of docker container linking.

bwatson-applause:
Hey Keycloak community!

This PR contains updates to attach containers via [docker-compose](https://docs.docker.com/compose/overview/) instead of relying on manual, multi-step docker container linking. Some benefits:

- All configuration for the 'cluster' is in one place.
- One, smaller command to bring up the 'cluster': `docker-compose up`
- The `server-ha-progress` can rely upon the `docker-compose.yml` file for command line modifications, instead of using an extra image or relying on the arguments being manually passed to `docker run`.
- More control over env var naming, rather than relying on the auto-generated env vars (e.g., `POSTGRES_PORT_5432_TCP_PORT`).

I also added an HAproxy instance in front of the Keycloak instances in the `server-ha-progres` subproject.

___IMPORTANT! Until the GA `jboss/keycloak-postgres` Image on Docker Hub is Updated___

Before running the `server-ha-postgres` example, you will need to build the image in `server-postgres` locally to prevent the existing `jboss/keycloak-postgres` from being pulled from Docker Hub:
```
docker build -t jboss/keycloak-postgres server-postgres
```
If the GA image is updated, this will no longer be an issue.

I look forward to your comments!

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,6 +5,8 @@ This is a set of Docker images related to Keycloak.
 - [keycloak](https://registry.hub.docker.com/u/jboss/keycloak/) example Keycloak server
 - [keycloak-mysql](https://registry.hub.docker.com/u/jboss/keycloak-mysql/) example connecting Keycloak to MySQL
 - [keycloak-postgres](https://registry.hub.docker.com/u/jboss/keycloak-postgres/) example connecting Keycloak to PostgreSQL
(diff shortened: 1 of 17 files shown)


bwatson-applause:
Having issues squashing the merge... closing and recreating.

PR #47, opened 2016-04-05
Use docker-compose instead of docker container linking.

bwatson-applause:
Hey Keycloak community!

This PR contains updates to attach containers via [docker-compose](https://docs.docker.com/compose/overview/) instead of relying on manual, multi-step docker container linking. Some benefits:

- All configuration for the 'cluster' is in one place.
- One, smaller command to bring up the 'cluster': `docker-compose up`
- The `server-ha-progress` can rely upon the `docker-compose.yml` file for command line modifications, instead of using an extra image or relying on the arguments being manually passed to `docker run`.
- More control over env var naming, rather than relying on the auto-generated env vars (e.g., `POSTGRES_PORT_5432_TCP_PORT`).

I also added an HAproxy instance in front of the Keycloak instances in the `server-ha-progres` subproject.

___IMPORTANT! Until the GA `jboss/keycloak-postgres` Image on Docker Hub is Updated___

Before running the `server-ha-postgres` example, you will need to build the image in `server-postgres` locally to prevent the existing `jboss/keycloak-postgres` from being pulled from Docker Hub:
```
docker build -t jboss/keycloak-postgres server-postgres
```
If the GA image is updated, this will no longer be an issue.

I look forward to your comments!

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,6 +5,8 @@ This is a set of Docker images related to Keycloak.
 - [keycloak](https://registry.hub.docker.com/u/jboss/keycloak/) example Keycloak server
 - [keycloak-mysql](https://registry.hub.docker.com/u/jboss/keycloak-mysql/) example connecting Keycloak to MySQL
 - [keycloak-postgres](https://registry.hub.docker.com/u/jboss/keycloak-postgres/) example connecting Keycloak to PostgreSQL
+- [keycloak-mongo](https://registry.hub.docker.com/u/jboss/keycloak-mongo/) example connecting Keycloak to MongoDB
 - [keycloak-adapter-wildfly](https://registry.hub.docker.com/u/jboss/keycloak-adapter-wildfly/) builds on top of the [jboss/wildfly](https://registry.hub.docker.com/u/jboss/wildfly/) image, adding the Keycloak adapter for Wildfly to it, as well as the required changes to the standalone.xml
 - [keycloak-examples](https://registry.hub.docker.com/u/jboss/keycloak-examples/) builds on top of [keycloak](https://registry.hub.docker.com/u/jboss/keycloak/), adding the examples.
 
+Additionally, this project contains [an example of a Keycloak HA cluster](server-ha-postgres).
--- a/server-ha-postgres/Dockerfile
+++ b/server-ha-postgres/Dockerfile
@@ -1,3 +1,0 @@
-FROM jboss/keycloak-postgres:latest
-
-CMD ["-b", "0.0.0.0", "--server-config", "standalone-ha.xml"]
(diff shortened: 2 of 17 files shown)


stianst:
Thanks for the PR. Afraid I'm going to reject it. We do not want to tie these images to Docker compose as there are several other solutions for managing orchestrating Docker images. We do not want to pick one and we don't have the resources to maintain multiple.

bwatson-applause:
Thanks for the speedy response! Do you mind if my company maintains and publishes these images?

xied75:
@bwatson-applause I guess being called official image means it should come from the "origin". And @stianst is correct to insist they serve the most common scenario which is probably one layer down the docker-compose level. To be frank docker-compose is not really used in production except local host testing, I believe you guys also have orchestrators like Kubernetes and friends. Hijacking the official image to fit what docker-compose demands is like forcing everyone else take that route, which no other official images (being nginx or postgre) would do.

bwatson-applause:
I think we have a miscommunication over my poorly worded request 😄 Let me clarify...

I'm not proposing hijacking the official images, trying to create a different official repo, or trying to force @stianst to accept these updates. I respect that this is the official repo for Keycloak Docker images, and that @stianst is moderating this repo. I respect his decision to reject my PR, and agree with his reasons for doing so (I would also state that it wouldn't matter if I disagreed, as it's his repo to moderate 😄).

When I asked "Do you mind if my company maintains and publishes these images?", I wasn't referring to a take over of the `jboss-dockerfiles/keycloak` repository. I was referring to transferring ownership of __my__ fork to my company, having my company keep that fork up-to-date with this repo, and publishing __those__ images under our namespace in Docker Hub. I'd also like to state I'm not trying to compete with this repository. I'm just trying to help spread Keycloak's adoption, as I have been impressed with it. I agree that docker composition isn't a production solution, as such deployments shouldn't have the service and data store running in the same instance. I was just thinking that offering an out-of-the-box `docker-compose up` solution that allows would-be adopters to see how easily Keycloak can be clustered behind a proxy would go far in helping others adopt Keycloak as an auth solution. That was the reason I created this fork in the first place.

However, I didn't want to just go off on my own, create, and publish those images without giving @stianst the chance to offer his input into my idea. Given that, @stianst, I look forward to your thoughts on my idea.

Thank you both for your time. I'm glad to see this project has such passionate contributors.

stianst:
@bwatson-applause that'd be very welcome. Send me a link once it's up as I plan to at some point create a page or a blog post listing external resources related to Keycloak. Also, thanks for being very understanding.

bwatson-applause:
Great, thank you! I will let you know when I have it up. Would you like me to add that info to this PR, or is there another communication channel you would like me to use?

stianst:
Adding a comment here will do

PR #49, opened 2016-05-11
Example Keycloak runnning behind an SSL reverse proxy

tonswieb:
Hi,

To be able to run the Keycloak example behind an Nginx reverse proxy with SSL offloading I had to manually make some changes to standalone.xml otherwise HTTPS url's were not rendered correctly.

I put this manual changes to standalone.xml in a separate example folder.
Perhaps it could be incorporated in the main example? I am assuming it does not lead to any conflict.

Regards,

Ton Swieb

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,4 +7,4 @@ This is a set of Docker images related to Keycloak.
 - [keycloak-postgres](https://registry.hub.docker.com/u/jboss/keycloak-postgres/) example connecting Keycloak to PostgreSQL
 - [keycloak-adapter-wildfly](https://registry.hub.docker.com/u/jboss/keycloak-adapter-wildfly/) builds on top of the [jboss/wildfly](https://registry.hub.docker.com/u/jboss/wildfly/) image, adding the Keycloak adapter for Wildfly to it, as well as the required changes to the standalone.xml
 - [keycloak-examples](https://registry.hub.docker.com/u/jboss/keycloak-examples/) builds on top of [keycloak](https://registry.hub.docker.com/u/jboss/keycloak/), adding the examples.
-
+- [keycloak-reverse-proxy](https://registry.hub.docker.com/u/jboss/keycloak-reverse-proxy/) example running Keycloak behind a reverse proxy
--- a/server-reverse-proxy/Dockerfile
+++ b/server-reverse-proxy/Dockerfile
@@ -1,0 +1,4 @@
+FROM jboss/keycloak:latest
+
+ADD changeProxy.xsl /opt/jboss/keycloak/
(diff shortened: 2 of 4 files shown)


stianst:
I can't accept this as we already have to many images. Would it be possible to do this with the main image and an environment variable instead?

tonswieb:
Hi Stian,

I think that is possible. I also prefer to have this in the main image. I
incorrectly assumed that you would be more reluctant to accept a separate
image then changes to the main image :-)

The socket-binding attribute is already parameterized I believe. So no
changes needed there, except perhaps some addition to the documentation.
The http-listener attribute proxy-address-forwarding="true" has not been
parameterized yet. But I believe that adding it does not have any negative
side-effects. As long as the X-Proto headers are not set then it will have
no effect.

If you agree then I would like to make a new pull request which adds
proxy-address-forwarding="true" to the http-listener and some extra
documentation for change the redirect-https to 443 using the existing
property.

Regards,

Ton


Met vriendelijke groet,

Ton Swieb
Adviseur Integratie

06 14680172
088 217 08 15
088 217 08 00
ton.swieb@finalist.nl

2016-05-12 6:09 GMT+02:00 Stian Thorgersen <notifications@github.com>:

> I can't accept this as we already have to many images. Would it be
> possible to do this with the main image and an environment variable instead?
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/jboss-dockerfiles/keycloak/pull/49#issuecomment-218655652>
>

-- 
 <http://www.finalist.nl>

stianst:
I assume you'd add proxy-address-forwarding="true" with the default as false and it being parameterized?

tonswieb:
I could do that as well. I thought adding it with the default as true
instead of false, because I do not see any negative side effects.

Met vriendelijke groet,

Ton Swieb
Adviseur Integratie

06 14680172
088 217 08 15
088 217 08 00
ton.swieb@finalist.nl

2016-05-20 7:03 GMT+02:00 Stian Thorgersen <notifications@github.com>:

> I assume you'd add proxy-address-forwarding="true" with the default as
> false and it being parameterized?
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/jboss-dockerfiles/keycloak/pull/49#issuecomment-220517875>
>

-- 
 <http://www.finalist.nl>

stianst:
It has security implications. Enabling proxy-address-forwarding results in the x-forwarded-for header being used. If there is no proxy or firewall in front of the Keycloak server in this setup an attacker can include a random value for the x-forwarded-for to for example bypass the brute force protection.

tonswieb:
Ok. Good to know. Then I will create a PR with proxy-address-forwarding
disabled by default.

Met vriendelijke groet,

Ton Swieb
Adviseur Integratie

06 14680172
088 217 08 15
088 217 08 00
ton.swieb@finalist.nl

2016-05-20 9:34 GMT+02:00 Stian Thorgersen <notifications@github.com>:

> It has security implications. Enabling proxy-address-forwarding results in
> the x-forwarded-for header being used. If there is no proxy or firewall in
> front of the Keycloak server in this setup an attacker can include a random
> value for the x-forwarded-for to for example bypass the brute force
> protection.
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/jboss-dockerfiles/keycloak/pull/49#issuecomment-220537788>
>

-- 
 <http://www.finalist.nl>

xtremebiker:
+1 for this feature. Would be really nice to enable the reverse proxy integration based in environment variables. As @stianst, I believe this would fit the main keycloak image and should be disabled by default.

stianst:
I'm closing this PR as it's not been updated. To request adding these features to the main Docker image please create an issue at https://issues.jboss.org/browse/KEYCLOAK. However, we would need someone from the community to contribute as we simply don't have the capacity at the moment.

PR #52, opened 2016-07-11
Add an example of setting keycloak up with SQL Server

realityforge:


stianst:
Thanks for your. I'm afraid I'm going to reject it as we don't have the capacity to maintain and test this image.

PR #54, opened 2016-09-01
Refactored POSTGRES environment variables

ragamiri:
The jdbc host address was set to POSTGRES_PORT_5432_TCP_ADDR. Updated it to POSTGRES_HOST and set the default to be the link name in the README (postgres).

The jdbc port was set to use env.POSTGRES_PORT_5432_TCP_PORT when the README specified POSTGRES_PORT as an environment variable.

These issues prevent keycloak from deploying.

### DIFF
--- a/server-postgres/README.md
+++ b/server-postgres/README.md
@@ -8,7 +8,7 @@ Extends the Keycloak docker image to use PostgreSQL
 
 First start a PostgreSQL instance using the PostgreSQL docker image:
 
-    docker run --name postgres -e POSTGRES_DATABASE=keycloak -e POSTGRES_USER=keycloak -e POSTGRES_PASSWORD=password -e POSTGRES_ROOT_PASSWORD=root_password -d postgres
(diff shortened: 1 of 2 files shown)


stianst:
This doesn't work as the value for env.POSTGRES_PORT is "tcp://172.17.0.2:5432" which is obviously not a valid port.

Stibbons:
There is no reason for `POSTGRES_PORT` to be different than 5432 if user set to. Please apply this patch.

stianst:
This patch breaks Postgres completely on my box. As I said env.POSTGRES_PORT resolves to "tcp://172.17.0.2:5432" which is not a valid port.

Stibbons:
indeed when using `--link`. For people using docker compose or similar (mesos for myself), we need a way to specify the port and host or the pg server.

stianst:
Feel free to create a JIRA and send a PR for that. However, it can't break things for others like this PR did.

PR #59, opened 2016-11-09
Set a reasonable JGroups bind default for keycloak-ha-postgres

solsson:
Unless you read https://keycloak.gitbooks.io/server-installation-and-configuration/content/v/2.3/topics/clustering/troubleshooting.html it is difficult to spot the fact that a ha setup using the `docker run` commands in https://hub.docker.com/r/jboss/keycloak-ha-postgres/ for 2.3.0.Final don't actually produce a running cluster.

To accept non-local cluster connections you have to set `-bprivate` or `-Djboss.bind.address.private`. Unlike the public and management ports, 0.0.0.0 is not a valid value here - see [this mailing list thread](https://sourceforge.net/p/javagroups/mailman/message/35478766/).

[Common wisdom](http://lists.jboss.org/pipermail/keycloak-user/2016-November/008273.html) seems to be to use `hostname -i`, which is non-trivial with docker args only. I therefore suggest that `docker-entrypoint.sh` for the `-ha` image exposes that as env.

With that in place and because the last parameter takes precedence we might just as well add it as argument to `/opt/jboss/keycloak/bin/standalone.sh`. An alternative would be to use `env.HOSTNAME_IP` inside the config file, but that makes overriding difficult.

I've also taken the liberty to rm standalone.xml, this being an image specifically for `-ha`. It's quite easy to forget about `--server-config` when you start adding your own command arguments, which reverts the image back to regular standalone behavior with logs saying "Adding singleton resource"). With the file removed you get `FileNotFoundException: /opt/jboss/keycloak/standalone/configuration/standalone.xml` instead.

Due to the warning "Avoid using this option; use hostname --all-ip-addresses instead." on `-i` in the man page for `hostname` there's an alternative env var `HOSTNAME_IP_ALL` which isn't used, but might be useful for additional args or config file changes.

### DIFF
--- a/server-ha-postgres/Dockerfile
+++ b/server-ha-postgres/Dockerfile
@@ -1,3 +1,5 @@
 FROM jboss/keycloak-postgres:latest
 
 CMD ["-b", "0.0.0.0", "--server-config", "standalone-ha.xml"]
+
+ADD docker-entrypoint.sh /opt/jboss/
--- a/server-ha-postgres/Dockerfile
+++ b/server-ha-postgres/Dockerfile
@@ -3,3 +3,5 @@ FROM jboss/keycloak-postgres:latest
 CMD ["-b", "0.0.0.0", "--server-config", "standalone-ha.xml"]
 
 ADD docker-entrypoint.sh /opt/jboss/
+
+RUN rm keycloak/standalone/configuration/standalone.xml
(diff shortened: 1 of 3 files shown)


solsson:
Closing this because #62 looks like a more reliable approach, with JGroups bound to the nic and JDBC_PING used to exchange IPs.

arvinep:
This works great,

Just a little change, we need to `chmod +x docker-entrypoint.sh` in order to avoid going to this error:

~~~
container_linux.go:262: starting container process caused "exec: \"/opt/jboss/docker-entrypoint.sh\": permission denied"
docker: Error response from daemon: oci runtime error: container_linux.go:262: starting container process caused "exec: \"/opt/jboss/docker-entrypoint.sh\": permission denied".
ERRO[0000] error waiting for container: context canceled
 ~~~

Otherwise, looks good.

PR #62, opened 2016-11-29
Add keycloak-ha-mysql with jgroups jdbc_ping

solsson:
This replaces my earlier https://github.com/jboss-dockerfiles/keycloak/pull/59, as I find MySQL/MariaDB a better fit for HA keycloak, in particular with Kubernetes:
 - MySQL is easy to cluster, thanks to Galera. See for example https://github.com/Yolean/kubernetes-mysql-cluster
 - JGroups JDBC_PING works out of the box with MySQL, but not with Postgres.

KUBE_PING (http://jgroups.org/manual4/index.html#_kube_ping, https://github.com/jgroups-extras/jgroups-kubernetes/) is an alternative, but the jars are not bundled by default and the configuration is complex. JDBC is easy to configure because JGroups can use the JNDI datasource from Keycloak.

Other suggestions in this PR:
 * A sample kubernetes setup in its own namespace, [server-ha-mysql/kubernetes-example.yml](https://github.com/Reposoft/keycloak-ha-kubernetes/blob/keycloak-ha-mysql/server-ha-mysql/kubernetes-example.yml)
 * Set a node identifier to get rid of txn warn, 2e1893f and d2d04a3
 * Suggest JGroups TCP instead of UDP in kubernetes. Our tests in GKE found that tcp generates fewer warnings and fewer errors at instance loss. 7d138fb
 * A shell script to facilitate more transformations like `changeDatabase.xsl` in the existing builds. `RUN xsl-transform ...` in `Dockerfile`s.


solsson:
Rebased on current, i.e. 2.5.1

solsson:
We've had this setup running in k8s production for a couple of months now. Occasionally we have issues with new instances being unable to join the cluster. Getting timeouts in jgroups. We resolve those manually by scaling down the cluster to 2 or sometimes 1 instances. If anyone who's proficient with jgroups reads this: Now every instance gets a new name. Would it help if we reused instance names when containers are rescheduled?

chexxor:
I am working on setting up KeyCloak on Kubernetes on GKE, and your branch worked great! I am running an off-cluster MySQL instance, though, so I just removed that definition from your yaml file and adjusted the connection details. Also, I needed to change `MYSQL_USER` to `MYSQL_USERNAME` - perhaps this is because I'm trying to use KeyCloak 3.0.0? Also, in the service definition, I forward port 80 to 8080 inside the keycloak pods.

Would be nice to have docs for the port defs - https://github.com/Reposoft/keycloak-ha-kubernetes/blob/keycloak-ha-mysql/server-ha-mysql/kubernetes-example.yml#L114-L139

solsson:
Confirmed, I changed to `MYSQL_USERNAME` in 1043ea9 as config says so. The readme was only half updated.

The ports are just my way to name all the ports from standalone-ha.xml (`kubectl -n keycloak exec kc-[something] -- cat /opt/jboss/keycloak/standalone/configuration/standalone-ha.xml`) while experimenting. Most of them can be removed now, but because this is just an example setup I figured others may want to experiment too.

I'd use the services to remap ports. In minikube I use `minikube service -n keycloak keycloak` to test.

c3s4r:
What's missing for this to be merged in master? Can I safely use your branch?

chexxor:
I believe maintainers aren't merging because they can't take ownership of it, perhaps because they aren't familiar with kubernetes.

c3s4r:
Thanks for your response @chexxor. Do you know if this is ready for Keycloak 3? I'd like to use it for production for my app but I don't know how stable or tested it is.

chexxor:
@c3s4r Yeah, I'm using it with Keycloak 3. I don't have heavy load from users, but it seems to work.

c3s4r:
Ok, great, I will give it a try and let you know how it goes. Thanks!

solsson:
@c3s4r We haven't upgraded to `:3` in production yet, but it looks ok in QA. Looking forward to hearing your results.

c3s4r:
Hi, I was testing this branch (modified to work with Keycloak 3.2.0.CR1). I tested locally and on Google Container Engine (GKE). It works great, but I have a few warnings:

1. When I scale up, sometimes it displays a warning like:

> 21:14:28,935 WARN  [org.jgroups.protocols.TCP] (TransferQueueBundler,ee,kc-4214536727-xcspj) JGRP000032: kc-4214536727-xcspj: no physical address for 91f96445-826d-f3b6-9ee3-ce16dfb25358, dropping message

Anyways, I can see in the database that the entry is correctly added in the JGROUPSPING table.

2. When I scale down, it displays this error on the instance that's being shut down: 

> 16:16:01,968 ERROR [org.jgroups.protocols.JDBC_PING] (MSC service thread 1-6) Could not open connection to database: java.sql.SQLException: javax.resource.ResourceException: IJ000470: You are trying to use a connection factory that has been shut down: java:jboss/datasources/KeycloakDS

And 20 times this error on the instances that are running:

> 21:13:43,049 ERROR [org.jgroups.protocols.TCP] (TransferQueueBundler,ee,kc-4214536727-xcspj) JGRP000029: kc-4214536727-xcspj: failed sending message to kc-4214536727-rgf69 (98 bytes): java.net.SocketTimeoutException: connect timed out, headers: RequestCorrelator: id=200, type=RSP, id=163, rsp_expected=false, FORK: ee:keycloak, UNICAST3: DATA, seqno=295, conn_id=3, TCP: [cluster_name=ee]

Anyways, the entry in the JGROUPSPING table is removed properly. This seems to be related to this issue in wildfly https://issues.jboss.org/browse/WFLY-6599?_sscc=t (which would be fixed in wildfly 11).

I don't know if this would be something to worry about. (Probably someone with more knowledge on JGROUPS and JDBC_PING can advise).

3. If you want to deploy in GKE, the name of the deployment should be really short. (I'm using kc for it to work). If I use "keycloak", I get the following error:

> 21:37:06,503 INFO  [org.jboss.as.clustering.jgroups] (ServerService Thread Pool -- 38) WFLYCLJG0001: Activating JGroups subsystem.
21:37:06,533 ERROR [org.jboss.as.controller.management-operation] (ServerService Thread Pool -- 49) WFLYCTL0013: Operation ("add") failed - address: ([("subsystem" => "transactions")]) - failure description: "WFLYCTL0112: 'keycloak-4214536727-46cpf' is an invalid value for parameter node-identifier. Values must have a maximum length of 23 characters"

I'm hoping that someone will take ownership of this, so we can have an official mysql-ha docker :) ... until that I will work with my own un-official branch.

Richie765:
Very useful branch. Got it finally working mostly. I added a few things to the Dockerfile, perhaps you could add it to this branch?

```dockerfile
# http://www.keycloak.org/docs/3.2/server_installation/topics/clustering/load-balancer.html
RUN sed -i -e 's/<http-listener /& proxy-address-forwarding="${env.PROXY_ADDRESS_FORWARDING}" /' $JBOSS_HOME/standalone/configuration/standalone-ha.xml

# http://www.keycloak.org/docs/3.2/server_installation/topics/cache/replication.html
ENV CACHE_OWNERS 1
RUN sed -i -e 's!<distributed-cache name="sessions" mode="SYNC" owners="1"/>!<distributed-cache name="sessions" mode="SYNC" owners="${env.CACHE_OWNERS}"/>!; s!<distributed-cache name="loginFailures" mode="SYNC" owners="1"/>!<distributed-cache name="loginFailures" mode="SYNC" owners="${env.CACHE_OWNERS}"/>!' $JBOSS_HOME/standalone/configuration/standalone-ha.xml
```

solsson:
@Richie765 I was under the impession that PROXY_ADDRESS_FORWARDING env was supported already.

Great addition with CACHE_OWNERS. Probably explains why we've occasionally lost sessions. I can add it when I update for 3.3.0.Final.

Richie765:
Great, thanks! I'm pretty sure `PROXY_ADDRESS_FORWARDING` wasn't supported already. It is set in the inherited base Keycloak Dockerfile, but that one applies to `standalone.xml`, not `standalone-ha.xml`.

solsson:
Maybe it's time I merge [keycloak3-ha-myqsl](https://github.com/Reposoft/keycloak-ha-kubernetes/tree/keycloak3-ha-mysql) to this branch? I guess the PR will be long lived.

In production we actually build from scratch these days, because a 200MB image makes re-scheduling across nodes faster than with the 600MB one from this repo. There I had already added PROXY_ADDRESS_FORWARDING. If this PR will live forever, it could be an option to maintain a repo that isn't a fork.

solsson:
Maintaners ( @stianst, @davidfestal ?), are you interested in maintaining an official keycloak-ha-mysql? If not, I will instead create a separate repo with only that image. It makes little sense to keep on rebasing this branch, as no actual code is shared between the different folders.

For the sake of this discussion, here's the self-contained build I referred to above: https://gist.github.com/solsson/7500f9e7140d3b37e8fe8b5f354139e5

stianst:
Hi, sorry that I'm late to the party :)

It would nice to have good examples for HA as well as Kubernetes example. A few thoughts:

* We've just introduced support for MySQL and PostgreSQL directly in Keycloak server image. So we don't need to have separate HA images for MySQL and PostgreSQL. See https://github.com/jboss-dockerfiles/keycloak/blob/master/server/README.md.
* We've added a specific OpenShift image. This probably needs some work, at least it's not enabling clustering at the moment. See https://github.com/jboss-dockerfiles/keycloak/tree/master/server-openshift
* We're working on changing to using jboss-cli instead of xml transformations as it's more future proof

With that in mind is this really a HA example or is it a Kubernetes example? This PR could be respun to keycloak-kubernetes or keycloak-ha. I'm in favor of keycloak-kubernetes as that allows introducing features specific to running Keycloak on Kubernetes.

stianst:
Main server image now updated to use jboss-cli instead of XML transformation. It also makes changes to both standalone.xml and standalone-ha.xml.

solsson:
I agree that a new repo keycloak-kubernetes would be the right scope. The choice of database has little to do with the choice of JGroups ping. Would you like to host such a project, and/or would you like me to set up one? Using PRs in this repo is impractical because it falls behind as the images are maintained (unless manifests and testing of them are owned here).

It's good news for keycloak-kubernetes that there'll be an image that contains drivers for both supported databases, because IMO all kubernetes examples that require an image build that isn't already on Docker Hub struggle with the separation. Docker Compose has built in support for building, while in Kubernetes it is specific to your choice of hosting.

That said, threre'll probably be tweaks to standalone-ha.xml in such a project, which gives us two options:
 * Maintain as standalone-ha.xml alongside k8s manifests, mount in the Keycloak pods (using ConfigMap probably)
 * A kubernetes specific image must be maintained in this repo and pushed to hub.docker.com/r/jboss/.

Your endorsement (or lack thereof) of a new Dockerfile is thus required anyway :)

stianst:
Let's go for a new image, keycloak-kubernetes, then. Once it's ready we can get arrange to have it released on Docker Hub. Bear in mind at least initially it will have to be maintained by the community as we don't have resources to maintain it.

solsson:
Ok. I can give it a shot, probably after 3.3.0.Final has been released. Is there more changes to the `jboss/keycloak` in the works right now? Is the transition to jboss-cli completed?

Richie765:
I've been using this branch a while with plain Docker (no Kubernetes). I don't know anything about k8s or what the k8s specifics in this image are. Will the future keycloak-kubernetes image still work without k8s? I've no plans for running k8s in the future, so it would be great if I could still use it somehow.

stianst:
@solsson Nothing more planned for now. I don't know when 3.3.0.Final will be released as we're waiting for WildFly 11 Final to be released.

stianst:
@Richie765 there's not that much here of value for you then? Isn't it just about using standalone-ha instead of standalone? Or do you rely on JDBC_PING as well?

Richie765:
I'm really a noob with Java and Jboss, but I'm pretty sure the regular keycloak-mysql isn't working in a cluster even when started with the standalone-ha.xml config. It starts alright but the nodes don't see each-other so don't replicate the caches.

I Googled around I found this image, since the changes here are in the keycloak-ha-mysql directory, I assumed it was what I needed. So I tried and it worked! So I'm guessing it is JDBC_PING that made the difference.

Is there anything really specific to k8s in this PR? If not, perhaps just call it keycloak-ha-mysql so it is clear it can be used with and without. Then everybody can benefit from this image.

stianst:
@Richie765 It may just be the fact that the "clustering" interface is bound to a loopback interface by default.

@solsson is there actually anything Kubernetes specific here? Could this not be keycloak-ha image instead? Then just have the Kubernetes compose files in a separate "kubernetes-examples" directory?

solsson:
I was actually thinking the same thing. There's the support for a [JGROUPS_STACK](https://github.com/Reposoft/keycloak-ha-kubernetes/pull/1/files#diff-955a616981152d6dcc36993d2c6e57b4R14) and [CACHE_OWNERS](https://github.com/Reposoft/keycloak-ha-kubernetes/pull/1/files#diff-52e9df99494c7bcb38bd13279fd5317fR13) env, but that could certainly be a `jboss/keycloak-ha`.

Besides, Kubernetes networking and/or JGroups behavior might have changed since the inception of this PR, so we should probably start out with testing a default ha setup once again. No contributor has explicitly reported that it doesn't work with the stock image (though I suspect the interest for this PR hints that it still doesn't).

I guess that introduction of new env vars, with current hard coded values turning into defaults, in standalone-ha.xml is a quite good way to maintain a generic ha image.

stianst:
For kubernetes-examples or compose-examples? there could be yaml files for a variant of things. MySQL + couple KC nodes, Postgres + couple KC nodes, KC + db + examples, etc.. So seems more correct to put those separate to the images themselves.

solsson:
I agree with the yaml strategy. I'm positive to having a generic keycloak-ha, and my reasoning went: introducing a bit more configurability-by-env will satisfy that strategy. Instead of adding to the confusion here, I will try to find the time to produce a new PR, if no one else beats me to it :)

Richie765:
Awesome that this is being worked on. I have a few theme related additions that you may want to consider for the image.

```dockerfile
# Set welcome theme
# http://www.keycloak.org/docs/3.2/server_development/topics/themes.html
# I've set it to redirect like this: https://stackoverflow.com/a/5411601/3653251
ENV WELCOME_THEME keycloak
RUN sed -i -e 's!<theme>!<theme>\n                <welcomeTheme>${env.WELCOME_THEME}</welcomeTheme>!' $JBOSS_HOME/standalone/configuration/standalone-ha.xml
```

```dockerfile
# Disable template caching, useful during theme development
# http://www.keycloak.org/docs/3.2/server_development/topics/themes.html
# I wanted to make a single TEMPLATE_CACHE for this but didn't know how
# Values need to be set to -1, false, false respectively to disable caching
ENV STATIC_MAXAGE 2592000
ENV CACHE_THEMES true
ENV CACHE_TEMPLATES true
RUN sed -i -r 's!<staticMaxAge>[[:digit:]]+</staticMaxAge>!<staticMaxAge>${env.STATIC_MAXAGE}</staticMaxAge>!' $JBOSS_HOME/standalone/configuration/standalone-ha.xml
RUN sed -i -r 's!<cacheThemes>[[:alpha:]]+</cacheThemes>!<cacheThemes>${env.CACHE_THEMES}</cacheThemes>!' $JBOSS_HOME/standalone/configuration/standalone-ha.xml
RUN sed -i -r 's!<cacheTemplates>[[:alpha:]]+</cacheTemplates>!<cacheTemplates>${env.CACHE_TEMPLATES}</cacheTemplates>!' $JBOSS_HOME/standalone/configuration/standalone-ha.xml
```

It can also be documented that a theme can be volume-mounted to `/opt/jboss/keycloak/themes/<theme>`

stianst:
@Richie765 Feel free to create a JIRA and send a PR to keycloak/server image for the themes options. I'd prefix the ENV varaibles with THEME_ though. Also, addition to keycloak/server/README.md on how to add a custom theme through volumne would be nice.

c3s4r:
@solsson do you have any updates? :)

solsson:
@c3s4r I gave it a shot a few days ago, but honestly I felt blocked by the switch to jboss-cli and the entrypoint script updating the config. I should give it more time, but since you asked... Porting the xsl is just a matter of time, but I have more of an issue with the entrypoint. I've come to appreciate the docker convention that config files are edited at image build time, not container run time.

The actual issue I had was that [this env trick](https://github.com/jboss-dockerfiles/keycloak/pull/62/files#diff-a77d3eece909d73f95a0b90990f73473R147) no longer fixes the issue with jdbc trying to connect to the pod's IP. There is probably better solutions to that problem, but the issue illustrates that with config modified at runtime I no longer know when the env vars have effect, and if the container dies on start it's nontrivial to know which config it is actually running.

Apologies if the above sounds like a rant. I think Yolean will keep using the standalone build that I mentioned [earlier](https://github.com/jboss-dockerfiles/keycloak/pull/62#issuecomment-333359371). With it we run keycloak's bin script directly.

I'm still prepared to do this, but my major purpose would be to re-test with Keycloak 3.3.0 and Wildfly 11, together with Kubernetes 1.8. It's a good opportunity to see if JGroups or GCP networking has changed. Thus I will wait for the final release of 3.3 and for GCP to allow 1.8 clusters.

c3s4r:
Thanks for the response @solsson . Don't know if you knew but GCP is providing early access to Kubernetes 1.8 ;) https://cloudplatform.googleblog.com/2017/09/google-container-engine-kubernetes-18.html (the link to enroll is at the bottom of the post) . I'd really like to have this image ready for a project that we're about to launch to production soon. I'd like to help, and I have plans to work on this within a couple of weeks. Let me know if you do some progress, so we can collaborate.

solsson:
@c3s4r What's your choice of database?

c3s4r:
We are using Google Cloud, because of that we will use mysql. (By now they are providing high availability and replication only on mysql).

solsson:
Good timing: http://wildfly.org/news/2017/10/23/WildFly11-Final-Released/ + https://cloud.google.com/container-engine/release-notes#october-24-2017 :)

solsson:
@c3s4r @stianst @Richie765 the new #92 is on par with this one now feature-wise. Let's continue the discussion there.

server-ha-mysql/jgroups-jdbc.xsl
```diff
@@ -0,0 +1,49 @@
...
+    </xsl:template>
+
+    <xsl:template match="/domain:server/domain:socket-binding-group/domain:socket-binding[@name='jgroups-mping']">
+        <xsl:comment> jgroups-mping was removed at switch to JDBC_PING </xsl:comment>
+    </xsl:template>
+
+    <xsl:template match="/domain:server/domain:interfaces/domain:interface[@name='private']">
+        <xsl:copy>
+            <xsl:apply-templates select="@*"/>
+            <nic xmlns="urn:jboss:domain:4.0" name="eth0"/>
```
mozesmagyar: something like this lets you specify the interface later
```
<nic xmlns="urn:jboss:domain:4.0" >                                                                    
    <xsl:attribute name="name">${jboss.bind.device.private:eth0}</xsl:attribute>                   
</nic>
```

PR #63, opened 2016-11-29
Make keycloak image OpenShift compatible (run as arbitrary uid)

jimmidyson:


jimmidyson:
Ping @stianst @goldmann

stianst:
I'm not convinced this image should be made OpenShift compatible as there's a lot more concerns with getting the image to run properly on OpenShift, such as clustering, setup of https, etc, etc.. Does it not also need templates and such to run, not just a Docker image?

jimmidyson:
Those things aren't strictly necessary: you can run `oc new-app jboss/keycloak` & it will deploy to your cluster.

You are right that for a production deployment you want templates, https config, etc, but you still need an image that runs as an arbitrary uid... I expect everything else to be done via a deployer.

These changes don't change existing behaviour of this image, just makes it possible to run directly on openshift. What are your concerns about incorporating these changes into the image so that others (me, hopefully, if I can find time) can build a deployment model for keycloak on openshift origin?

stianst:
What do you mean by "everything else to be done via a deployer"? Bear in mind by OpenShift v3 knowledge is pretty poor..

As OpenShift aka Kubernetes makes it rather simple to scale an instance clustering should be supported outright I think.

In general we need a community story for Keycloak as we only have one for RH-SSO at the moment, but I don't think this covers it. There's much more to it, so I don't want to have a partial fix here now to then get people complaining and requesting this that and the other. I'd rather have a proper and documented approach to OS v3 online and origin.

jimmidyson:
My thinking is that the image is just the runtime. Configuration, provision of certificates, default realms/clients, etc are separate concerns but things like in this PR can't be changed by the deployer.

stianst:
I appreciate that, but as we are planning to have a proper community image in the future (early next year) I'd rather not start adding bits and pieces that may break the image for non-OpenShift users.

I'm happy to keep the PR open for now, but I don't have time or anyone else that has time to review it right now. I also don't know yet if our OpenShift support will use this image or another image.

jimmidyson:
There is nothing in here that would break it for non-OpenShift users btw...

mohammedzee1000:
@jimmidyson @stianst @goldmann I have done the investigation required to validate that basic function of container is not affected.
The log is too big to paste here so im giving a pastebin for the same instead

http://pastebin.centos.org/64081/

server-postgres/Dockerfile
```diff
@@ -2,5 +2,5 @@ FROM jboss/keycloak:latest
 
 ADD changeDatabase.xsl /opt/jboss/keycloak/
 RUN java -jar /usr/share/java/saxon.jar -s:/opt/jboss/keycloak/standalone/configuration/standalone.xml -xsl:/opt/jboss/keycloak/changeDatabase.xsl -o:/opt/jboss/keycloak/standalone/configuration/standalone.xml; java -jar /usr/share/java/saxon.jar -s:/opt/jboss/keycloak/standalone/configuration/standalone-ha.xml -xsl:/opt/jboss/keycloak/changeDatabase.xsl -o:/opt/jboss/keycloak/standalone/configuration/standalone-ha.xml; rm /opt/jboss/keycloak/changeDatabase.xsl
-RUN mkdir -p /opt/jboss/keycloak/modules/system/layers/base/org/postgresql/jdbc/main; cd /opt/jboss/keycloak/modules/system/layers/base/org/postgresql/jdbc/main; curl -O http://central.maven.org/maven2/org/postgresql/postgresql/9.3-1102-jdbc3/postgresql-9.3-1102-jdbc3.jar
+RUN mkdir -p /opt/jboss/keycloak/modules/system/layers/base/org/postgresql/jdbc/main; cd /opt/jboss/keycloak/modules/system/layers/base/org/postgresql/jdbc/main; curl -O https://repo1.maven.org/maven2/org/postgresql/postgresql/9.3-1102-jdbc3/postgresql-9.3-1102-jdbc3.jar
```
stianst: Why are you changing this?

server/Dockerfile
```diff
@@ -9,17 +9,23 @@ USER root
...
 
-USER jboss
-
 RUN cd /opt/jboss/ && curl -L https://downloads.jboss.org/keycloak/$KEYCLOAK_VERSION/keycloak-$KEYCLOAK_VERSION.tar.gz | tar zx && mv /opt/jboss/keycloak-$KEYCLOAK_VERSION /opt/jboss/keycloak
 
 ADD docker-entrypoint.sh /opt/jboss/
 
 ADD setLogLevel.xsl /opt/jboss/keycloak/
 RUN java -jar /usr/share/java/saxon.jar -s:/opt/jboss/keycloak/standalone/configuration/standalone.xml -xsl:/opt/jboss/keycloak/setLogLevel.xsl -o:/opt/jboss/keycloak/standalone/configuration/standalone.xml
+RUN sed -i 's/\(logger.handlers=CONSOLE\),FILE/\1/' keycloak/standalone/configuration/logging.properties
```
stianst: Why are you changing this?

server/Dockerfile
```diff
@@ -9,17 +9,23 @@ USER root
 
 RUN yum install -y epel-release && yum install -y jq && yum clean all
 
-USER jboss
```
stianst: Why are you moving this?

server/Dockerfile
```diff
@@ -9,17 +9,23 @@ USER root
...
 
 ADD docker-entrypoint.sh /opt/jboss/
 
 ADD setLogLevel.xsl /opt/jboss/keycloak/
 RUN java -jar /usr/share/java/saxon.jar -s:/opt/jboss/keycloak/standalone/configuration/standalone.xml -xsl:/opt/jboss/keycloak/setLogLevel.xsl -o:/opt/jboss/keycloak/standalone/configuration/standalone.xml
+RUN sed -i 's/\(logger.handlers=CONSOLE\),FILE/\1/' keycloak/standalone/configuration/logging.properties
 
 ENV JBOSS_HOME /opt/jboss/keycloak
 
+RUN mkdir $JBOSS_HOME/standalone/{data,log} && \
```
stianst: Why are you changing this?

server/docker-entrypoint.sh
```diff
@@ -4,5 +4,4 @@ if [ $KEYCLOAK_USER ] && [ $KEYCLOAK_PASSWORD ]; then
     keycloak/bin/add-user-keycloak.sh --user $KEYCLOAK_USER --password $KEYCLOAK_PASSWORD
 fi
 
-exec /opt/jboss/keycloak/bin/standalone.sh $@
-exit $?
+exec /opt/jboss/keycloak/bin/standalone.sh $@
```
stianst: Why are you changing this? Adding the exit statement was added for a reason.

server/setLogLevel.xsl
```diff
@@ -17,6 +17,10 @@
         </xsl:copy>
     </xsl:template>
 
+    <xsl:template match="//log:handler[@name='FILE']" />
```
stianst: Why are you changing this?

stianst:
I'm reluctant to add all these changes at the moment and there are quite a few changes here. If you remove all unnecessary changes and explain the changes that are absolutely needed I may be convinced to merge it.

We are also not sure what the strategy will be for creating images for OpenShift. It may or may not rely on this image. If it doesn't rely on this image all these changes are not needed.

AFAIK the images produced by cloud enablement team doesn't even rely on DockerFiles, so there's a good chance these images won't be part of the OpenShift images at all.

server-postgres/Dockerfile
jimmidyson: Downloading from non-https site isn't really a good idea, especially when a https version is available. Just an extra security thing I guess.

server/Dockerfile
jimmidyson: To remove the need to log to file in the container. Only stdout/stder (CONSOLE) logging should be enabled. Saves on disk for redundant files.

server/docker-entrypoint.sh
jimmidyson: `exec` system call replaces the current shell with the specified program so exit will never actually be called.

server-postgres/Dockerfile
stianst: Missed the change in scheme. Ok, so this change is fine.

stianst: It's completely separate change to OpenShift support though and would have been much easier to merge as a separate PR.

server/Dockerfile
stianst: I'd argue the opposite. It's important to log to file. This change has nothing to do with OpenShift support though.

mohammedzee1000:
@stianst AFAIK dockerfiles is just one of the easiest way of producing the containers. As long as whatever you use to build the image produces an the same in the format that can be consumed by an OCI compatible runtime, it will run as a container.

server/Dockerfile
jimmidyson: This block sets up necessary writable dirs to be able to run as arbitrary user. In an OpenShift deployment, the run as user is set to arbitrary uid based on the project uid range. The `chmod`s below set up stuff as required to be readable/writable by anyone in root group (not a security risk btw). See https://docs.openshift.com/enterprise/3.2/creating_images/guidelines.html#use-uid for more details.

server/setLogLevel.xsl
jimmidyson: Again, to remove the FILE logger which in a container is redundant as we should log to stdout/stderr.

server/docker-entrypoint.sh
stianst: True, this change is fine. It does have nothing to do with OpenShift though.

server/Dockerfile
jimmidyson: Why do you want to log to a file when in a container? stdout/stderr streams are captured by the execution engine (i.e. Docker). You're right this has nothing to do with OpenShift, so don't mind removing it if it at least allows a community image to run on OpenShift, but while I was updating this file thought I'd tidy up a few other things too.

stianst:
I assumed that all proposed changes where related to OpenShift and not general improvements. That's one reason why I was rather reluctant to add these changes.

To move forward please break your changes into separate PRs and also create JIRA issues (https://issues.jboss.org/projects/KEYCLOAK) for each proposed changed. Explain why it's needed. That allows us to review and merge changes individually and also allows others to view what has changed and gives us history of changes that may be needed to debug regressions if they occur.

I'm going to close this PR as it is simply covering to many different things in one go.

jimmidyson:
Um, thanks?!?! I send in a PR that helps your project run on a strategic platform, then offered to remove the changes unrelated to OpenShift from this PR to do the minimal changes to get this to run on OpenShift & you close the PR because I haven't raised JIRAs, an undocumented requirement to contribute to this repo?

stianst:
I'm not trying to be difficult here, but please appreciate that at the moment we have other higher prioritizes. OpenShift is going to be a priority for us going forwards, but as I've said I do not know yet if this image will even be part of that strategy. Nor do I know when we aim to provide community images for OpenShift.

I welcome all your proposed improvements, including the non-OpenShift changes. I do not welcome bundling many changes together in one PR though. I think you can appreciate that makes it very difficult to review and merge the proposed changes. This is common to most projects and very few will accept contributions that bundle many unrelated changes together.

Raising JIRA first is not undocumented. Please take a look at our website where it clearly explains this. I appreciate that it's not directly mentioned in this repo though, so I understand why you where unaware of this. However, please appreciate that we do need to track changes that was made. There's nothing novel about having a issue tracking system for an open source project.

I'm sorry if you feel that your contribution has not been received well and I admit I could have dealt with this better.

mohammedzee1000:
@stianst This image is already being used by the opensource projects under Redhat Developers. As of the moment the images being used are are from my fork where the changes have been applied. I had to go ahead as it required as part of summit goals.

stianst:
@mohammedzee1000 please get in touch with me on email and let's discuss details there. It could also be helpful to setup a call.

@jimmidyson same goes to you, please get in touch with me on email to discuss what your needs are and how we can proceed.

PR #65, opened 2017-02-24
Added option to enable modcluster multicast listens on host's ip inst…

jmowla:
modcluster multicast listens over "private" interface which is defaulted to loopback ip (127.0.0.1)
To enable modcluster to listens on reachable IP address from other containers within network we need to send "jboss.bind.address.private"

To make this happens "jboss.bind.address.private" variable needs to be set to host's IP address, so there are two changes:
1. Dockerfile to sends extra variable
2. docker-entrypoint.sh to processes $(hostname -i) correctly. (this file will be updated during the build)

### DIFF
--- a/server-ha-postgres/Dockerfile
+++ b/server-ha-postgres/Dockerfile
@@ -1,5 +1,9 @@
 FROM jboss/keycloak-postgres:latest
 
-RUN sed -i 's/\$@/\$\(eval "echo \$@"\)/' /opt/jboss/docker-entrypoint.sh
+WORKDIR /opt/jboss
+ADD src/NetInfo.java .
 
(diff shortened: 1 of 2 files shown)


solsson:
How about binding to the interface instead, as in https://github.com/jboss-dockerfiles/keycloak/pull/62 https://github.com/jboss-dockerfiles/keycloak/pull/62/commits/67483e6c9bd6aaa2ec477b197388dd473c1a5536? Worked for us both in standalone docker and in Kubernetes. No need for NetInfo code.

stianst:
I really don't like this approach. It's rather messy using a Java class to find this info. There's plenty of linux tools already that can do this. Also why not just bind to interface as suggested by @solsson.

stianst:
Closing due to lack of feedback to comments

PR #66, opened 2017-03-02
Add standalone-ha modifications

bpalermo:


stianst:
This only gets you halfway there as it'll still use standalone.xml. Wouldn't https://github.com/jboss-dockerfiles/keycloak/pull/73 could do both.

stianst:

PR #67, opened 2017-03-16
Fix jboss history

rassie:


stianst:
Thanks for your PR, but I'm going to reject it. New releases of Keycloak won't come with Mongo support so there won't be any updated versions of the Mongo image. Check our blog for more details.

rassie:
That's a shame, but understandable. Thank you nevertheless.

PR #70, opened 2017-04-27
Making Proxy address forwarding configurable

secondsun:
I've added the ability to configure proxy address forwarding and documented how to do this in the README.

The easiest way to test it is to set a SSL edge proxy and see if the Admin Console link on the /auth page links to an http or https address.  This is easiest to set up in OpenShift as an edge terminated route.


stianst:
Thanks - added https://issues.jboss.org/browse/KEYCLOAK-4815

PR #71, opened 2017-05-04
Fix proxy address forwarding layer

dkelleher:
Move the `sed` command below the definition of `JBOSS_HOME`.  It appears
standalone.xml doesn't exist at the point when the `sed` was run.
Additionally the environment variable `JBOSS_HOME` was not defined.

Also added the equals sign to the `server/README.md`. Without an equals
sign docker assumes true is the container name which causes it to fail.


dkelleher:
I should also ask is this worth moving to an xsl file like the other config changes use?

stianst:
Ideal would be JBoss CLI in offline mode.

PR #72, opened 2017-05-15
Support Arbitrary User IDs so keycloak starts on openshift

rawlingsj:


rawlingsj:
Just tested and this works on both:
minikube version: v0.19.0
minishift version: 1.0.1

mguillem:
+1

we are using our custom Keycloak image since some time due to the same issue:
```
FROM jboss/keycloak:2.5.4_Final

USER root
RUN chmod -R o+rw /opt/jboss/keycloak
```

secondsun:
A very similar PR was merged a few weeks ago : https://github.com/jboss-dockerfiles/keycloak/pull/68

PR #73, opened 2017-06-16
Add env. variable for modified file

klinki:
We need to run wildfly with `standalone-full.xml` configuration, so I had to create custom Dockerfile.
It would be nice if configuration file could be specified in variable.


stianst:
Could you rebase please?

stianst:
Also would make sense that the env variable is used to determine what config is executed as well.

stianst:
Closing due to lack of feedback. Please reopen once comments are addressed.

PR #74, opened 2017-06-19
Add SAML support to wildfly OIDC adapter

cainj13:
Wanted a wildfly container that included the SAML adapter, as I did not see one out there.  Went ahead and implemented.

Note that I'm currently on KC 2.5.5, so that's the version I used for this.  May want to add some tags/change for a broader coverage of versions.

Have a working test of this here: [https://hub.docker.com/r/joshcain/jboss-dockerfiles-keycloak/](https://hub.docker.com/r/joshcain/jboss-dockerfiles-keycloak/), where I've been able to successfully deploy and secure SAML applications.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 .idea
 *.iml
+*.swp
--- a/adapter-oidc-saml-wildfly/Dockerfile
+++ b/adapter-oidc-saml-wildfly/Dockerfile
@@ -1,0 +1,13 @@
+FROM jboss/keycloak-adapter-wildfly:2.5.5.Final
+
+ENV KEYCLOAK_VERSION 2.5.5.Final
+
+WORKDIR /opt/jboss/wildfly
+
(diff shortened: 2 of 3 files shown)


cainj13:
Hi all - just wanted to ping on this.  Is this something we're going to get in?

Starting to use this container more broadly for some rapid dev purposes, but if it's not going to land here, then I'm probably going to stand up a more official docker account and host it there.

stianst:
If both adapters can con-exist without problems I don't see any reason why this couldn't just be added to adapter-wildfly image. As we don't support 2.x in community anymore I'm not going to update old tags, but would only add changes to next releases of the images.

cainj13:
@stianst Sure.  See #75 .

Only thing I can think of is that if people are acting really container-y and doing things like only putting a single app in each container, you'd never actually *need* both to live side-by-side simultaneously.  Would just add a little bloat in most cases...

stianst:
I don't see the bloat. It's just a few kb discspace. IMO it's better to have more capabilities in one image than an explosion of images.

cainj13:
Sure, if that's where the team lands on the tradeoff, I'm fine either way.  We should be all set for #75 then?

stianst:
Closing in favour of #75

PR #75, opened 2017-07-07
add SAML adapter to wildfly image

cainj13:
Per @stianst 's comment on #74 , just adding the SAML adapter to the same wildfly server as the OIDC adapter.

Going to leave the old PR out there - some folks that only use one or the other might not like having the additoinal subsystems installed on the wildfly server...


cainj13:
bump.  Any thoughts?

cainj13:
@stianst any more thoughts here?

stianst:
@cainj13 been on holiday last 4 weeks. Will take a look soon.

cainj13:
@stianst  bump.

stianst:
Thanks for the bump ;)

PR #77, opened 2017-08-09
Correct documented default password used for MySQL

fedde-s:
The password used by default to access a MySQL database does not match
the one listed in the README file. Correct the README file.


fedde-s:
Ah, I just noticed that the Postgres setup has the same issue. Those should really be fixed in the same commit.

fedde-s:
Pull request #78 fixes both README files in one commit.

PR #79, opened 2017-08-25
Make `keycloak-openshift` multi-database and compatible with S2I.

davidfestal:
Signed-off-by: David Festal <dfestal@redhat.com>


davidfestal:
@stianst Is there anything more to do to have this PR reviewed merged ?

davidfestal:
@stianst @goldmann could anyone review this PR please ? This work is related to the need to use a `keycloak-postgres` on Openshift Online along with a `s2i` build.

stianst:
Thanks for your PR. Not sure about this one to be honest. What I wanted to do was to add MySQL and PostgreSQL support directly to the main image to prevent having so many images. That's especially important with these OpenShift specific images.

stianst:
Will need to discuss this with the team and find out what our strategy with regards to community OpenShift images should be

davidfestal:
> Thanks for your PR. Not sure about this one to be honest. What I wanted to do was to add MySQL and PostgreSQL support directly to the main image to prevent having so many images. That's especially important with these OpenShift specific images.
>
> Will need to discuss this with the team and find out what our strategy with regards to community OpenShift images should be

Do you think this would be possible to merge this PR first and then start thinking about some global refactoring regarding databases and openshift support ? And when things get clearer, the already-existing `keycloak-postgres-openshift` image could then be deprecated or simply be changed to be based on the new refactored image ?

That would allow us going further with our own project, without having to store this image in some other location.

stianst:
Could you send me an email and describe what project you need this for? I'd rather continue this discussion over email.

davidfestal:
@stianst I reworked this PR in order to:
- push back into the main image the files that are required to change the database, so that they are available for derived images
- enhance the `keycloak-openshift` image which wasn't published until now, so that is is S2I-compatible, and supports the 3 database modes (embedded, mysql or postgres).

> What I wanted to do was to add MySQL and PostgreSQL support directly to the main image to prevent having so many images. That's especially important with these OpenShift specific images.

That's mainly what is done in this PR now: though the behavior of already-published images (`keycloak`, `keycloak-mysql`, `keycloak-postgres`, ...) is still exactly the same as before from a user POV, there is now only one openshift-compatible image that derives from the main `keycloak` image and supports all the databases.

I hope this will allow you merging this PR more easily.

stianst:
Nice. Thanks for doing this!

One suggestion: could we make the DB selection more automatic? Instead of having the different startup scripts the single startup script could detect what DB to use based on whatever first matches for:

1. Is DB_VENDOR env value set then use that
2. Is there any env values prefixed with MYSQL_ - then use MySQL
3. Is there any env values prefixed with POSTGRES_ then use Postgres
4. Fallback to H2

Also, would it be problematic to just add the DB work directly to the main Keycloak image rather than the OpenShift specific one?

stianst:
> I'm not a fan of implicitly guessing the DB based on environment variable prefixes honestly.

Why not? It would make it much simpler to use. I can't see why MYSQL_ env value would be set if MySQL isn't used, nor the other way around where MYSQL_ env value is set but it's not used. Seems perfectly legit approach to me.

davidfestal:
> Why not? It would make it much simpler to use. I can't see why MYSQL_ env value would be set if MySQL isn't used, nor the other way around where MYSQL_ env value is set but it's not used. Seems perfectly legit approach to me.

OK, why not. I can test for `POSTGRES_DATABASE` or `MYSQL_DATABASE`.

stianst:
Thanks, if you do have some concerns about that approach please share though ;)

davidfestal:
@stianst I made the last changes you proposed (main change is [here](https://github.com/jboss-dockerfiles/keycloak/pull/79/files#diff-d0463d5d7460c28597180b6a29b001b8R8) ).

I updated the Readme of the `keycloak-openshift` image, but didn't touch to the documentation of the main image, since the default behavior is still the same as before.

And I prefer you to do the documentation change on the main image : I don't really know how you want to present this, or if you plan to remove the `keycloak-postgres` and `keycloak-mysql` images, or keep them and point from the main image documentation to each one according to the chosen database.

stianst:
Nice work, thanks for doing this :)

I can update the server README file. My plan is to keep MySQL and PostgreSQL images for now, but just have them set the DB_VENDOR env and point folks to use the main server image instead.

davidfestal:
Thanks @stianst for merging. This morning I found the updated `keycloak-openshift` image in docker hub, great !

However could you explain me what is required to have the image tagged with a version number such as `3.0.0.CR2` in the docker hub ? Currently only the `latest` tag is available, but I'm not sure using a "floating" tag such as `latest` would be acceptable for our project.

stianst:
I don't have access to creating automatic builds on Docker Hub directly and
the guy who added it for me forgot to give me permissions to control the
settings for the build. Once he's added that I'll create a tag.

3.2.0.CR2 tag OK? I can just update the tag for that in GitHub and make
sure it builds in Docker hub. FIY In the future we automatically create
tags when doing KC releases which makes Docker hub automatically build.

davidfestal:
`
Did you mean `3.3.0.CR2` ? it seems to be the version displayed when I select *Admin -> Server Info* in the Keycloak admin console.


`

stianst:
Indeed, that's what I meant ;)

stianst:
3.3.0.CR2-2 on it's way:
https://hub.docker.com/r/jboss/keycloak-openshift/builds/

I didn't want to change the existing 3.3.0.CR2 tag, so added -2 just for
you.

davidfestal:
Thanks very much ! it's perfect !

davidfestal:
@stianst Any idea why this build failed though:

https://hub.docker.com/r/jboss/keycloak/builds/bvbljsacbg6h8jpdaprxqqd/ ?

since `keycloak-openshift` is based on `keycloak`. I'm wondering if the `keycloak` build error could have had any impact on the quality of the `keycloak-openshift` image (when it was generated or when it is used now)

stianst:
Yes, I know why:
https://hub.docker.com/r/jboss/keycloak/builds/bvbljsacbg6h8jpdaprxqqd/

There's no Keycloak version 3.3.0.CR2-2.

Should be fixed now.

PR #80, opened 2017-08-30
Change namespace version for datasources

rbriois:
Datasources namespace in Keycloak configuration files is : 
```xml
<subsystem xmlns="urn:jboss:domain:datasources:5.0">
```

The .xsl files used by Dockerfile to customize Keycloak configuration file still use 4.0 version :
```xml
<xsl:stylesheet version="2.0"
                xmlns:xsl="http://www.w3.org/1999/XSL/Transform"
                xmlns:ds="urn:jboss:domain:datasources:4.0">
```

This PR correct the problems by updating namespace versions in .xsl files


stianst:
Thanks. We need to get this changed to use jboss-cli so this doesn't happen again.

PR #82, opened 2017-09-06
Make Postgres JDBC version configurable

ernsheong:

PR #83, opened 2017-09-06
Fix error in retrieving Postgres JDBC module

ernsheong:
Apologies, fix error introduced by https://github.com/jboss-dockerfiles/keycloak/pull/82

Related JIRA: https://issues.jboss.org/browse/KEYCLOAK-5418

PR #84, opened 2017-09-07
Fixes issue introduced by PR #82, bad jdbc driver file naming

trecloux:
With pull request #82, we end up with this situation:
```
[jboss@9c2cf66b4e19 ~]$ ls -l /opt/jboss/keycloak/modules/system/layers/base/org/postgresql/jdbc/main
total 704
-rw-r--r-- 1 root  root    1354 Sep  6 09:33 module.xml
-rw-r--r-- 1 jboss jboss      0 Sep  6 09:35 postgres-jdbc.jar
-rw-r--r-- 1 jboss jboss 713037 Sep  6 09:35 postgresql-42.1.4.jar
```


trecloux:
This is also fixed in PR #83

stianst:
Merged #83

PR #86, opened 2017-10-03
KEYCLOAK-4803 Moved support for MySQL and PostgreSQL to main Keycloak…

stianst:
… server image


solsson:
Note that this affects [export using bin/standalone.sh](http://www.keycloak.org/docs/latest/server_admin/index.html#_export_import) through custom entrypoint or exec. You'll get a freshly initialized H2 db. There's a workaround in the commit above.

PR #90, opened 2017-10-24
Fix postgres container environment variables

Toilal:
According to [postgres official image docs](https://hub.docker.com/_/postgres/), environment variable for database name is POSTGRES_DB instead of POSTGRES_DATABASE, and no ROOT_POSTGRES_PASSWORD is available.


stianst:
Thanks

PR #91, opened 2017-10-25
Specify default name for PostgreSQL host

rstub:
The documentation states that one should start a PostgreSQL container and link it to the Keycloak container with name postgres. However, it seems this name is never used. Specifying the environment variable POSTGRES_PORT_5432_TCP_ADDR works for me, but the above change should get the documented approach working.


stianst:
Could you paste how you start the Postgres image and how you start Keycloak image? The linking works just fine for me

rstub:
Hmm, it indeed works when starting the containers individually. I had used `docker-compose` and forgot about the link environment variables being only supported with v1 compose files (https://docs.docker.com/compose/link-env-deprecated/). Do you have any plans to replace the link environment variables?

stianst:
We don't at the moment, but feel free to create a JIRA and send a PR. At a
quick glance at the page you linked it looks like we should.

On 26 Oct 2017 9:40 am, "Ralf Stubner" <notifications@github.com> wrote:

Closed #91 <https://github.com/jboss-dockerfiles/keycloak/pull/91>.

—
You are receiving this because you commented.

Reply to this email directly, view it on GitHub
<https://github.com/jboss-dockerfiles/keycloak/pull/91#event-1311472590>,
or mute the thread

PR #92, opened 2017-10-29
keycloak-ha image with Kubernetes example

solsson:
... and JGroups TCP ping support.

Supercedes https://github.com/jboss-dockerfiles/keycloak/pull/62.

No README updates yet. I will document only in commit comments and this PR, until the structure has been reviewed by others.


solsson:
It'll be a mix of jboss-cli, sed and xsl for now, until we know what the desired end result is, at which point I hope someone can help me with jboss-cli.

solsson:
I fail to reuse the JDBC_PING transform from the previous PR: https://github.com/Reposoft/keycloak-ha-kubernetes/commit/93b80f88f438028177151a1a3c922f39eeeb9c9c

I think my quota of weekend work is up now, but you never know :)

solsson:
I've confirmed that neither `udp` nor `tcp` stacks JGroups works out of the box in a GKE cluster.

As in #62 `JDBC_PING` works. Unlike there, I fail to use an env to select stack in standalone-ha.xml. Error:
```
14:53:11,615 ERROR [org.jboss.as.controller] (Controller Boot Thread) WFLYCTL0362: Capabilities required by resource '/subsystem=jgroups/channel=ee' are not available:
    org.wildfly.clustering.jgroups.channel-factory.${env.JGROUPS_STACK:udp}; Possible registration points for this capability: 
		/subsystem=jgroups/channel=*
		/subsystem=jgroups/channel=*/fork=*
		/subsystem=jgroups/stack=*
```
The workaround a133254 is unsuitable for a Dockerfile.

I haven't tried [KUBE_PING](https://github.com/jgroups-extras/jgroups-kubernetes) this time around.

solsson:
I'm done hacking now. @stianst I get the feeling that everything here could be part of _the_ server image, since you're heading that direction anyway. We could simply describe the added env vars in a HA section in the readme.

 * `CACHE_OWNERS` with default `1` is backwards compatible.
 * `JGROUPS_STACK` with default `udp`, assuming we find the issue with a133254, is also backwards compatible.
 * For Keycloak, or any JGroups app with a JDBC connection, I think it makes sense to default to JDBC_PING instead of MPING. There might be environments that block UDP but support multicast TCP, but to get there with the docker image you would have edited the config file anyway to get the tcp stack. Thus the change seems low-risk.
   - if we do this we should also remove the jgroups-mping socket-binding.
 * 60ff485 looks like it fixes an omission, since `jboss.node.name` is an existing var.

@Richie765 Thanks again for contributing CACHE_OWNERS. Regarding https://github.com/jboss-dockerfiles/keycloak/pull/62#issuecomment-334088236 I agree with @stianst it's a separate PR.

Richie765:
@solsson, you're welcome, thanks for all the work on this repo as well, it's been very useful. I'm planning to make another PR for the templating things later. I worked a lot on the templates lately and have quite a few additions I can contribute. Only I'm a bit hesitant because I did a PR on another repo not long ago and the review process was very frustrating to me. But I'll do it anyway.

patriot1burke:
Instead of doing all these XLST transformations, why not a standalone.xml file that is preconfigured that comes with the Keycloak distro?

stianst:
That'd make a lot of sense. We could add standalone-ha-aws|kubernetes|openshift files to the distro.

solsson:
@patriot1burke Of course, but to encourage experiments here I had to make it a script on startup. Not everyone can rebuild docker images easily in their clusters, while this approach gives you a few seconds dev roundtrip with `kubectl apply -f kubernetes-example.yaml` followed by `kubectl scale` down and up.

This PR is currently in the stage where we try to determine what changes such a standalone-ha-* (kubernetes in my case) would have. The choice of xsl, sed or jboss-cli is just an implementation detail. @stianst I'm very interested in your feedback on that. In https://github.com/jboss-dockerfiles/keycloak/pull/92#issuecomment-340553216 I forgot to list https://github.com/jboss-dockerfiles/keycloak/pull/92/commits/61c244fc0779e7e46734b956859ead3257547d9e).

stianst:
Using the yaml file to do the configuration is a good idea for now. Eventually most (if not all) should be in the main Docker image (or a docker-kubernetes?) image.

I'd like to see it use kube ping though as that should work better in Kubernetes than JDBC ping. I know some folks have had issue with JDBC ping, but not sure about the details.

I'd also like to see it use jboss-cli instead of xsl/sed or whatever. jboss-cli is a bit of a pita to understand, but it's a lot more future proof and easier to maintain than xsl as they tend to not work properly when we upgrade WF. It's also quite hard to review this as it refers to external xsl files (can't merge it this way that's for sure).

stianst:
The yaml file should be moved to 'kubernetes-examples' directory in the root instead of 'keycloak-ha' dir.

solsson:
> I'd like to see it use kube ping

Agreed. I've evaluated the new build now and it has the same issue as #62: records are not removed from the jgroups table when pods are deleted. The cluster still works, but there's lots of ERROR log entries about failing to connect to peers.

> I'd also like to see it use jboss-cli

Agreed :) I started using it now and the validation with jboss-cli is a huge advantage compared to xsl/sed.

@stianst Are there any plans to upgrade to JGroups 4 with Keycloak? JDBC_PING has matured quite a bit, but 1.0+ is incompatible with current JGroups, see https://github.com/jgroups-extras/jgroups-kubernetes/issues/30#issuecomment-309672292.

I've started to investigate KUBE_PING in #94. It's really hard to find any resources on how to use it, so I think JDBC_PING wins if I also have to go back to a jgroups-kubernetes version that the authors no longer maintain.

solsson:
JGroups 4 added [DNS_PING](http://jgroups.org/manual4/index.html#_dns_ping), which looks adequate for Kubernetes and unlike KUBE_PING is included in [core](https://github.com/belaban/JGroups/commit/eb099c4a2cb946335714b7ae15e0f91594432a80#diff-820c741e3ab91b3967cd392b7ddcc1ef). It could probably even be enabled at runtime, on existence of some env, in the standard image much like database selection is handled today.

piwi91:
Hi @solsson,

I'm trying out your branch to setup a HA KeyCloak on K8S, but I'm not getting it to work. Don't know if it should with your latest example?

```
10:13:45,810 ERROR [org.jboss.as.controller.management-operation] (Controller Boot Thread) WFLYCTL0013: Operation ("add") failed - address: ([
    ("subsystem" => "datasources"),
    ("data-source" => "ExampleDS")
]) - failure description: {"WFLYCTL0288: One or more services were unable to start due to one or more indirect dependencies not being available." => {
    "Services that were unable to start:" => ["org.wildfly.data-source.ExampleDS"],
    "Services that may be the cause:" => [
        "jboss.concurrent.ee.tsp",
        "jboss.txn.ArjunaRecoveryManager",
        "jboss.txn.ArjunaTransactionManager",
        "jboss.txn.JBossContextXATerminator",
        "jboss.txn.TransactionManager",
        "jboss.txn.TransactionSynchronizationRegistry",
        "jboss.txn.UserTransaction",
        "jboss.txn.UserTransactionAccessControlService",
        "jboss.txn.UserTransactionRegistry",
        "jboss.txn.context.local"
    ]
}}
```

solsson:
@stianst I need to chose an approach to stabilize with, as a real docker build. Can you hint if JGroups 4 is in the roadmap or not?

If not I'll go for JDBC_PING, as it is more generic and has the advantage that instances sharing a database inherently share the ping, avoiding intricacies with labels etc. Will try to solve the purge issue.

Does "We're now testing database replication with MySQL Galera and Oracle RAC" (from 3.4.1 CR1 release notes) have implications here?

stianst:
We're not going to upgrade to JGroups 4 until WildFly does, not sure when that is.

It should be possible to add KUBE_PING to JGroups 3 as well though. I think that would be the preferred way.

Database replication within a single DC should be fine with regards to JDBC_PING, but if you have a cross DC setup then you can't rely on the same DB as Keycloak as that'll be replicated cross the DCs, but you don't want the nodes in the cluster to be replicated cross the DCs.

stianst:
There's:
https://search.maven.org/#artifactdetails%7Corg.jgroups.kubernetes%7Ckubernetes%7C0.9.3%7Cjar

Also:
https://github.com/jboss-openshift/openshift-ping

Both should work in JGroups 3

stianst:
@pedroigor Did you have a working clustered Keycloak docker image? Is it using Kube Ping?

solsson:
> We're not going to upgrade to JGroups 4 until WildFly does, not sure when that is.

Ok, that's background enough. Let's focus on JGroups 4.

> It should be possible to add KUBE_PING to JGroups 3 as well though. I think that would be the preferred way.

Sure, it's probably not that hard, but after my research I'm not too keen on trusting an old version of it in production, as there seems to be some caveats and to us it's not worth the risk.

I'll stick with JDBC_PING which has worked well enough and later if/when JGroups 4 makes it into Wildfly we'll likely switch to [DNS_PING](https://github.com/jboss-dockerfiles/keycloak/pull/92#issuecomment-345495202). It's simple, included in core and a great fit with how Kubernets services work.

> Database replication within a single DC should be fine with regards to JDBC_PING, but if you have a cross DC setup then you can't rely on the same DB as Keycloak as that'll be replicated cross the DCs, but you don't want the nodes in the cluster to be replicated cross the DCs.

Thanks. Sounds like a different use case, with more state going through the DB.

@stianst Are you at all interested in a JDBC_PING based image? If so I'll try to dockerize this PR properly, with jboss-cli, so we as a community can maintain it. If not I'll just prop up our own image built from debian and the Keycloak tar.

piwi91:
@solsson I'm interested in the JDBC_PING based image 😄

solsson:
I'm interested in sharing the maintenance with a community as well. I could set up a separate repo for the docker build instead. My experience from #62 is that it's hard to maintain a long-lived PR, because all tracking must be done in that github issue, while a separate repo can have proper issue tracking and releases.

stianst:
We'd be very interested in an HA image that works for Kerberos and OpenShift. Not sure it can be the same for those two. Not sure it should be JDBC_PING. Maybe we could make it easily configurable with an environment variable like we've done with the databases? Could just have different standalone-ha.xml files for the different configs that are picked based on the config option.

piwi91:
@solsson Sounds good to me and I agree it's better to create a new repository to focus more on a Kubernetes solution.

stianst:
For the record we are planning to provide a Kubernetes option and contributions around it would be great. The problem at the moment is that we can't focus on it until early next year as we are busy wrapping up the next RH-SSO release.

I can see two options here. Either continue working on this PR and wait until we have time to work with you. Or just setup a separate repo and hopefully we can align in the future.

With regards to ping option, I've seen some hints that maybe DNS_PING may be the way to go in the future, but I don't know to much about it. Hopefully in the new year I'll have time to look at Kubernetes more, if not someone else from the Keycloak team for sure will!

solsson:
@rayscunningham Great. This uses the runtime config concept from ./server, so we could easily add support for JGROUPS_PING too, if the value of the env var is read.

You might want to add a cli script for CACHE_OWNERS, see https://github.com/jboss-dockerfiles/keycloak/pull/62#issuecomment-333219621 and 860fb1e.

solsson:
I guess the takeaway here is that an image should support different kinds of JGroups setups. Hence I've created #100.

Please also consider #95.

@piwi91 I think #100 is as good as JDBC_PING gets, so feel free to try.

patriot1burke:
@solsson 
 You say that people cannot build their own images...BUT...If somebody needs to tweek themes or write an extension, extending the image will be a requirement.  I'm only half finished but I've created a separate "kubernetes" distro that includes all the KUBE_PING libraries, a standalone-kub.xml file.  I'm also working on a template quickstart that has k8s .yaml files to set up a keycloak cluster with an ingress and with session affinity.

Like Stian said, I'm still distracted with Keycloak 3.4.x and won't finish it up until the new year.

solsson:
@patriot1burke 
> You say that people cannot build their own images...

I only said "Not everyone can rebuild docker images easily in their clusters" to explain why I had done this PR as a script in a configmap. If we can agree on what to bundle with the official image, example kubernetes manifests can use those. In #100 I built the image locally and used `imagePullPolicy: Never` but I'm quite sure this will put some people off from testing.

On the contrary I recommend everyone to build their own images. I provided an example of our internal build in https://github.com/jboss-dockerfiles/keycloak/pull/62#issuecomment-333359371. But I think it's very valuable if we as a community can share `.cli` recipes etc, along with our experiences of clustering.

> I'm also working on a template quickstart that has k8s .yaml files to set up a keycloak cluster with an ingress and with session affinity.

Great! I'd be happy to remove the kubernetes-examples folder from #100. I think a Quickstart is a better scope than example manifests, because the latter comes with a lot of undocumented biases.

> includes all the KUBE_PING libraries

I'm only arguing that this isn't worth the libs. Ping is a relatively simple problem, as #100 demonstrates. KUBE_PING is a relatively complex solution, as the history of https://github.com/jgroups-extras/jgroups-kubernetes indicates.

PR #93, opened 2017-11-14
Fix typo from org.keycloack to org.keycloak

czeckd:
I would have filed this just as an issue, but issues aren't available for this project.


stianst:
Thanks

solsson:
@czeckd Did you notice any effect of changing this log level? I can't see any difference between `docker run --rm -e KEYCLOAK_LOGLEVEL=DEBUG -t jboss/keycloak:3.4.0.Final` and `docker run --rm -e KEYCLOAK_LOGLEVEL=WARN -t jboss/keycloak:3.4.0.Final`, not even on the occasional entries that are `[org.keycloak.services]`.

solsson:
#95 suggests something like the above

PR #95, opened 2017-11-20
Improve support for DEBUG level logging, including runtime level change

solsson:
The README includes an instruction for loglevel change, by env var, that has no effect because:
- Docker users only see stdout (not file logging)
- Not a lot of logging comes from `org.keycloak` package, which the env regulated. Most is from `org.jboss`.

In addition, runtime log level change is a great feature for troubleshooting. Keycloak supports that, so I documented a hint on how to use it.

### DIFF
--- a/server/README.md
+++ b/server/README.md
@@ -117,6 +117,8 @@ When starting the Keycloak instance you can pass a number an environment variabl
 
     docker run -e KEYCLOAK_LOGLEVEL=DEBUG jboss/keycloak
 
+Note that there are still per-package loglevel defaults.
+
 ### Enabling proxy address forwarding
 
 When running Keycloak behind a proxy, you will need to enable proxy address forwarding.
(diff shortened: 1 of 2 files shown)


stianst:
Removing the file logger is fine. The console logger shouldn't be set to anything the line "<level name="INFO"/>" should just be removed.

In many cases Keycloak debug is what you are interested in and only having the option to change everything or nothing is not good. I do appreciate that sometimes you want to change the log level for everything, but we should have a separate environment variable to set that.

What about something like this (removed unrelated bits):

     <subsystem xmlns="urn:jboss:domain:logging:3.0">
          <console-handler name="CONSOLE">
               <formatter>
                    <named-formatter name="COLOR-PATTERN"/>
               </formatter>
          </console-handler>
          <logger category="org.keycloak">
               <level name="${env.KEYCLOAK_LOGLEVEL:info}"/>
          </logger>
          ...
          <root-logger>
               <level name="${env.ROOT_LOGLEVEL:INFO}"/>
               <handlers>
                    <handler name="CONSOLE"/>
               </handlers>
          </root-logger>
          ...
     </subsystem>

stianst:
I need this now so I'm going to make the changes I suggested and prepare a new PR.

solsson:
Thanks. I had this on my todo-list but it didn't make it to the top. Well resolved in #121.

PR #96, opened 2017-12-06
Added Kubernetes JGroups KUBE_PING support

rayscunningham:
Created 'server-kubernetes' subfolder with Dockerfile to build with all the JGroups KUBE_PING module dependencies for JGroups 3.   It uses the jboss cli to configure.  I was able to successfully deploy and run in Google Kubernetes Engine (GKE) backed by Postgres DB.  Please see the server-kubernetes/README.md for additional info. 

Modified service/docker-entrypoint.sh for optionally enabling where 'JGROUPS_PROTOCOL=kubernetes'

if [ $JGROUPS_PROTOCOL ]; then
    echo "[KEYCLOAK DOCKER IMAGE] Using JGroups $JGROUPS_PROTOCOL"
    /bin/sh /opt/jboss/keycloak/bin/change-jgroups.sh $JGROUPS_PROTOCOL
fi

This has not been tested yet in any live Production environment.  I am leaving it to the community to help test this out further.


server-kubernetes/cli/jgroups/standalone-ha-configuration.cli
```diff
@@ -0,0 +1,3 @@
+embed-server --server-config=standalone-ha.xml --std-out=echo
+run-batch --file=cli/jgroups/${env.JGROUPS_PROTOCOL}/change-jgroups.cli
```
solsson: Good way to use the env var. I'm curious about the choice if env var name "_PROTOCOL". Is it more than the PING that needed changing? In my experience you only need to replace `tcp` `MPING` with something elsr.

piwi91:
@rayscunningham I would like to test it in our test environment. Do you already pushed a docker image to docker hub or other pub repository?

solsson:
Should have written here what I wrote in #92:

You might want to add a cli script for CACHE_OWNERS, see https://github.com/jboss-dockerfiles/keycloak/pull/62#issuecomment-333219621 and 860fb1e.

And have you managed to avoid the warnings about node-identifier? Otherwise you might also want to add a cli script for 60ff485.

rayscunningham:
@piwi91 I don't have a docker image on a public repository, but you can create an image from this pull request or the fork.  Build the 'server' folder with tag 'jboss/keycloak' first.   Then you can build the 'server-kubernetes' folder.

rayscunningham:
@solsson Thanks for the info.  I will be adding the CACHE_OWNERS and node-identifier to the cli script.

rayscunningham:
Added node-identifier, CACHE_OWNERS, and PROXY_ADDR_FORWARDING.  I updated the README with info on these.

Currently there appears to be a dependency issue when the CACHE_OWNERS > 1, I am seeing log statements complaining about missing dependencies such as "WFLYCTL0412: Required services that are not installed:" => ["jboss.concurrent.ee.tsp"].  I'll try to look at this further, whenever I have some more time.

rayscunningham:
There appears to be  a 23 character limit for the transactions node-identifier so changed to use POD_IP instead of POD_NAME.  No longer seeing depending errors now even with CACHE_OWNERS > 1.

solsson:
Yep. A different workaround is to name the deployment `kc`.

piwi91:
@rayscunningham I'm testing this PR and have some issues because I use RBAC. Any thoughts?
```
10:25:48,684 WARN  [org.jgroups.protocols.kubernetes.KUBE_PING] (ServerService Thread Pool -- 50) Problem getting Pod json from Kubernetes Client[masterUrl=https://10.10.0.1:443/api/v1, headers={}, connectTimeout=5000, readTimeout=30000, operationAttempts=3, operationSle
ep=1000, streamProvider=org.openshift.ping.common.stream.TokenStreamProvider@293b1ab2] for cluster [ejb], namespace [keycloak], labels [null]; encountered [java.lang.Exception: 3 attempt(s) with a 1000ms sleep to execute [OpenStream] failed. Last failure was [java.io.IOE
xception: Server returned HTTP response code: 403 for URL: https://10.10.0.1:443/api/v1/namespaces/keycloak/pods]]
```

simonpayne58:
@rayscunningham i have tested this from your original fork and it worked for me.  I built your image after having already pulled jboss/keycloak:latest.  I have yet to plan more thorough testing, but i observed i was able to scale the cluster as expected.

rayscunningham:
@stianst Do you have any comment or feedback about this pull request?

stianst:
@rayscunningham I need to find some time to play with Kubernetes before I can comment on this properly and review it. Hopefully in the next few week / month or so.

RBlaine95:
Can this work out of the box in an OpenShift Origin setup, or is there much configuration needed?  
I'm trying to get a HA Keycloak Cluster working in OpenShift Origin and this seems promising

server-kubernetes/jgroups/kubernetes/module.xml
```diff
@@ -0,0 +1,37 @@
...
+  ~ but WITHOUT ANY WARRANTY; without even the implied warranty of
+  ~ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+  ~ Lesser General Public License for more details.
+  ~
+  ~ You should have received a copy of the GNU Lesser General Public
+  ~ License along with this software; if not, write to the Free
+  ~ Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+  ~ 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+  -->
+<module xmlns="urn:jboss:module:1.5" name="org.jgroups.kubernetes">
```
rhusar: Changing the module.xml is very fragile and gets out of date. You can employ `module="org.jgroups.kubernetes"` on the protocol.

rayscunningham:
@RBlaine95 I don't know what would be involved with OpenShift Origin.  I've only tried on Kubernetes cluster on Google Cloud Platform.  You can give it a try with the configuration options mentioned in server-kubernetes/README.md

rhusar:
@andrejmaya  You need to grant the service account a view permission.

rhusar:
@andrejmaya Something like `oc policy add-role-to-user view system:serviceaccount:$(oc project -q):default -n $(oc project -q)`

andrejmaya:
@rhusar perfect thx! The cluster is running and the pods seem to find each other. As soon as I try to login, I get this error here:
```
00:39:52,272 WARN  [org.keycloak.events] (default task-3) type=CODE_TO_TOKEN_ERROR, realmId=Dkv, clientId=keycloak-sample-app, userId=null, ipAddress=10.0.1.12, error=expired_code, grant_type=authorization_code, code_id=1640e1f5-d640-4bc0-b94b-bddc7d8205f4, client_auth_method=client-secret
```
Seems that the code is not distributed across the cluster. Did someone experienced the same?

deitch:
Thanks @rhusar . I assume once it goes GA, it eventually will be used by keycloak, but assumptions have done in far greater people than I!

From the current `pom.xml` https://github.com/keycloak/keycloak/blob/master/pom.xml#L47 :

```xml
        <wildfly.version>11.0.0.Final</wildfly.version>
```

So it looks like currently `11.0.0.Final`.

Can anyone comment on future plans?

Also, I figure we will need some env vars in the entrypoint to configure it?

rhusar:
@deitch The only modification required is configuring the ping protocol to `kubernetes.KUBE_PING` and provide `KUBERNETES_NAMESPACE` and `KUBERNETES_LABELS` environmental variables. See https://github.com/jgroups-extras/jgroups-kubernetes for more info.

deitch:
Yeah I saw that. Won’t we need an env var on the keycloak docker image to tell it to use the kube_ping protocol?

stianst:
I will review this in detail soon. Hopefully next week. A few comments so far:

* This is related to https://github.com/jboss-dockerfiles/keycloak/pull/100. Rather than adding a new Kubernetes image we should add support for DNS_PING and KUBE_PING to the main image.
* We are working on upgrading to WildFly 13. Hopefully it will be ready for 4.0.0.Final. That make it much easier to add KUBE_PING

kjarnet:
Awesome - this seems to be exactly what we need to set up Keycloak clustered in our k8s setup! Have set this up in our test environment and seems to work perfectly so far (just minimal extra config required).

deitch:
> we will need something to change the ping protocol itself for k8s. This seems like its best handled by a docker-entrypoint script (in a similar fashion this PR is attempting to do

Yep, @rhusar , an env var or command-line option is the "standard" containerized way of doing this, so makes sense to have something that consumes it.

marvinosswald:
@rayscunningham I'm testing the setup right now, got the communication with k8s running but the don't seem to see each other on a new bootet pod i see the following while there are two more:
`Received new cluster view for channel ejb: [keycloak-7fc4f88c44-b76cp|0] (1) [keycloak-7fc4f88c44-b76cp]`

frekele:
Jgroups 4.0.x supports [DNS discovery](http://www.jgroups.org/manual4/index.html#_dns_ping) and [kubernetes](http://www.jgroups.org/manual4/index.html#_kube_ping).

Jgroups 3 has [KUBE_PING ](http://www.jgroups.org/manual/index.html#_kube_ping) support but it does not work very well, is very limited, in jgroups version 4.x this works very well with wildfly 13.

It's really missing the keycloak being compatible with kubernetes.

An upgrade to 'org.jgroups: jgroups: 4.0.13.Final' would be great.

This to be possible keycloak needs to be upgraded to WildFly 13.

There is already an ongoing task for this created on 01/Mar/2018:
https://issues.jboss.org/browse/KEYCLOAK-6781
That has been pushed forward for some time, since the '4.2.0.Final'

I think the update is being very complicated.

marvinosswald:
@frekele what do you mean by is very limited and not working very well ? Does it work or not ? It feels like i'm just an inch away from solving this for our current Keycloak setup as it all seems to work but the pods not seeing each other.

Any advice, hacky solution appreciated as i won't be able to wait for the implementation of Wildfly 13

frekele:
@marvinosswald 

In my tests and from other developers, we realized that wildfly 11+ jgroups 3.x does not work well on a unicast kubernetes cluster, some nodes are not properly discovery, it simply do not see them, already in a cluster baremetal multicast seems to work fine.
When in the cloud in the case of AWS we use S3_PING that works perfectly.

After switching to wildfly 13 + jgroups 4.x with 'DNS discovery' worked very well with kubernetes in unicast cloud and baremetal.

As keycloak still uses wildfly 11 and jgroups 3.x I believe the problem you are having is similar to what we had, but it depends on each infrastructure.

What infrastructure do you use?

marvinosswald:
@frekele yes i'm with aws as well, i hoped to use the clean solution but i guess i'm going to switch to s3 now as well. Just doesn't feel right.

frekele:
@marvinosswald 

I can tell you that s3_ping works very well with jgroups 3.x.

We have three networks.

1 baremetal for local development.
1 aws-cloud for homologation.
1 aws-cloud for production.

We are using wildfly 13 for our API with jgroups dns_ping for kubernetes.

For keycloak we are using keycloak 4.x with s3_ping in homologation and production.

Also we would like to use dns_ping or kube_ping with keycloak, but it does not seem to behave well, we are going to wait until we have a keycloak version compatible with the new wildfly 13 and jgroups 4.

Hope this helps.

marvinosswald:
To whom it may concern,
To who finds this thread,

and decides to use S3_PING as well.

S3_PING is only compatible with `AWS Signature Version 2` which is **not supported** by the following regions:
- US East (Ohio) Region
- Canada (Central) Region
- Asia Pacific (Mumbai) Region
- Asia Pacific (Seoul) Region
- EU (Frankfurt) Region
- EU (London) Region
- China (Beijing) Region

> Anything that can go wrong will go wrong
> -Edward Murphy

stianst:

stianst:

rhusar:
@marvinosswald Never use S3_PING. It's horrible and deprecated. Use NATIVE_S3_PING (https://github.com/jgroups-extras/native-s3-ping) instead. S3_PING will be removed in 5.0.

slaskawi:
Thanks a lot for the contribution! 

We decided to refactor the HA part a bit. In https://github.com/jboss-dockerfiles/keycloak/pull/151 we proposed a universal mechanism for using different discovery protocols. In the Pull Request we uploaded a test image so that you can play with it. Please have a look at the Pull Request and put your comments there.

@stianst Please close it in favor of https://github.com/jboss-dockerfiles/keycloak/pull/151

PR #97, opened 2017-12-13
docker setup for mysql

maozza:
add docker-compose.yml examples for MySQL.
fix for MySQL connection string + exposing MySQL host and port to environment variables.

### DIFF
--- a/server/README.md
+++ b/server/README.md
@@ -47,7 +47,7 @@ You can also use the `DB_VENDOR` environment variable to explicitly specify the 
 
 First start a MySQL instance using the MySQL docker image:
 
-    docker run --name mysql -e MYSQL_DATABASE=keycloak -e MYSQL_USER=keycloak -e MYSQL_PASSWORD=password -e MYSQL_ROOT_PASSWORD=root_password -d mysql
+    docker run --name mysql -e MYSQL_ADDR=mysql -e MYSQL_PORT=3306 -e MYSQL_DATABASE=keycloak -e MYSQL_USER=keycloak -e MYSQL_PASSWORD=password -e MYSQL_ROOT_PASSWORD=root_password -d mysql
 
 #### Start a Keycloak instance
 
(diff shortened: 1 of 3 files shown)


server/README.md
```diff
@@ -59,7 +59,7 @@ Start a Keycloak instance and connect to the MySQL instance:
 
 When starting the Keycloak instance you can pass a number of environment variables to configure how it connects to MySQL. For example:
 
-    docker run --name keycloak --link mysql:mysql -e MYSQL_DATABASE=keycloak -e MYSQL_USER=keycloak -e MYSQL_PASSWORD=password jboss/keycloak
+    docker run --name keycloak --link mysql:mysql -e MYSQL_ADDR=mysql -e MYSQL_PORT=3306 -e MYSQL_DATABASE=keycloak -e MYSQL_USER=keycloak -e MYSQL_PASSWORD=password jboss/keycloak
```
stianst: When using --link mysql:mysql passing the environment variables shouldn't be needed.

server/cli/databases/mysql/change-database.cli
```diff
@@ -1,5 +1,5 @@
 /subsystem=datasources/data-source=KeycloakDS: remove()
-/subsystem=datasources/data-source=KeycloakDS: add(jndi-name=java:jboss/datasources/KeycloakDS,enabled=true,use-java-context=true,use-ccm=true, connection-url=jdbc:mysql://${env.MYSQL_PORT_3306_TCP_ADDR}:${env.MYSQL_PORT_3306_TCP_PORT}/${env.MYSQL_DATABASE:keycloak}, driver-name=mysql)
+/subsystem=datasources/data-source=KeycloakDS: add(jndi-name=java:jboss/datasources/KeycloakDS,enabled=true,use-java-context=true,use-ccm=true, connection-url=jdbc:mysql://${env.MYSQL_ADDR:mysql}:${env.MYSQL_PORT:3306}/${env.MYSQL_DATABASE:keycloak}, driver-name=mysql)
```
stianst: This change breaks the --links functionality

victorperin:
@stianst depending on `links` is not a best practice.

For example: using docker-compose will it not work.

I think using `MYSQL_ADDR` and `MYSQL_PORT` seems the best way, since you already need to add user and password as env vars too.

victorperin:
Or could be updating the `change-database.cli` to verify if those variables exist. If not, use those ADDR and PORT vars.

stianst:
'--link' is removed sure, but we can't just remove it without warning in case folks rely on it. I finally took some time and look at thew new recommended 'network' way and it looks pretty nice. Working on a PR now that would support that as well as have legacy support for '--link'. Should be ready soon.

stianst:
Closing this PR in favour of https://github.com/jboss-dockerfiles/keycloak/pull/103.

I added your Docker Compose example and another one for PostgreSQL as well. There is a problem with the MySQL example as the MySQL database is not ready to receive connections when Keycloak starts. Seems to work fine on PostgreSQL. Would probably have to add some sort of check before starting Keycloak to make MySQL is running. Or maybe that can also be done through the datasource config.

PR #104, opened 2018-02-22
export POSTGRES_PORT

fisache:
For other port (not 5432) in postgres, the port number must be exported as env.


stianst:
I'm going to reject this as we've removed using the '--link' approach which has been deprecated in Docker for a while now.

PR #105, opened 2018-03-24
Add JDBC parameters as environment variable

thaleshonda:
Add a variable environment **POSTGRES_SSL** to specify ssl connections

### DIFF
--- a/docker-compose-examples/keycloak-mysql.yml
+++ b/docker-compose-examples/keycloak-mysql.yml
@@ -23,6 +23,7 @@ services:
         MYSQL_PASSWORD: password
         KEYCLOAK_USER: admin
         KEYCLOAK_PASSWORD: Pa55w0rd
+        JDBC_PARAMS: "connectTimeout=30000"
       ports:
         - 8080:8080
       depends_on:
--- a/docker-compose-examples/keycloak-mysql.yml
+++ b/docker-compose-examples/keycloak-mysql.yml
@@ -23,7 +23,8 @@ services:
         MYSQL_PASSWORD: password
         KEYCLOAK_USER: admin
         KEYCLOAK_PASSWORD: Pa55w0rd
-        JDBC_PARAMS: "connectTimeout=30000"
+        # Uncomment the line below if you want to specify JDBC parameters. The parameter below is just an example, and it shouldn't be used in production without knowledge. It is highly recommended that you read the MySQL JDBC driver documentation in order to use it. 
+        #JDBC_PARAMS: "connectTimeout=30000"
       ports:
         - 8080:8080
       depends_on:
--- a/docker-compose-examples/keycloak-postgres.yml
+++ b/docker-compose-examples/keycloak-postgres.yml
@@ -22,6 +22,7 @@ services:
         POSTGRES_PASSWORD: password
         KEYCLOAK_USER: admin
         KEYCLOAK_PASSWORD: Pa55w0rd
+        JDBC_PARAMS: "ssl=true"
       ports:
         - 8080:8080
       depends_on:
--- a/docker-compose-examples/keycloak-postgres.yml
+++ b/docker-compose-examples/keycloak-postgres.yml
@@ -22,7 +22,8 @@ services:
         POSTGRES_PASSWORD: password
         KEYCLOAK_USER: admin
         KEYCLOAK_PASSWORD: Pa55w0rd
-        JDBC_PARAMS: "ssl=true"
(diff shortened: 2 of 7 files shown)


rvion:
yes please !
current default is not ideal.

-----------

#### Possible improvement to this PR

📝 [this repo](https://github.com/mnasyrov/keycloak-postgres-ssl) adds `POSTGRES_JDBC_PARAMS` (and `MYSQL_JDBC_PARAMS`) 

```
docker run --name keycloak-postgres-ssl -p 8080:8080 \
    -e POSTGRES_JDBC_PARAMS=ssl=true&sslmode=verify-ca \
    ...
```

I think this is better because it's more powerfull and future proof.

Thanks anyway for the PR !

stianst:
+1 I also prefer the generic approach. We also need to update the readme file otherwise folks won't know this exists.

rbarilani:
@thaleshonda I did not find any support channel so I will try to ask here:

Do you ever seen this error?

```
Operation: /subsystem=datasources/data-source=KeycloakDS: add(jndi-name=java:jboss/datasources/KeycloakDS,enabled=true,use-java-context=true,use-ccm=true, connection-url=jdbc:postgresql://po
stgres:5432/keycloak, driver-name=postgresql)
Failure: WFLYCTL0097: Wrong type for 'connection-url'. Expected [EXPRESSION, STRING] but was OBJECT
```

docker-compose-examples/keycloak-mysql.yml
```diff
@@ -23,6 +23,7 @@ services:
         MYSQL_PASSWORD: password
         KEYCLOAK_USER: admin
         KEYCLOAK_PASSWORD: Pa55w0rd
+        JDBC_PARAMS: "connectTimeout=30000"
```
stianst: Not sure we should have this in the example. Can you elaborate on why you added it?

docker-compose-examples/keycloak-postgres.yml
```diff
@@ -22,6 +22,7 @@ services:
         POSTGRES_PASSWORD: password
         KEYCLOAK_USER: admin
         KEYCLOAK_PASSWORD: Pa55w0rd
+        JDBC_PARAMS: "ssl=true"
```
stianst: Not sure we should have ssl on by default. Rather just cover it in the readme.

server/docker-entrypoint.sh
```diff
@@ -73,6 +73,7 @@ else
     DB_NAME="embedded H2"
 fi
 
+export JDBC_PARAMS=$(echo ${JDBC_PARAMS} | sed '/^$/! s/^/?/')
```
stianst: Can you elaborate on what this is doing?

docker-compose-examples/keycloak-mysql.yml
thaleshonda: I think it would be useful include a JDBC parameter in the exemple, although I could not think in a better example than that. You think is better removing this parameter here?

docker-compose-examples/keycloak-postgres.yml
thaleshonda: Same from the MySQL example, I thought it would be useful include a parameter in the example. Should we remove it?

server/docker-entrypoint.sh
thaleshonda: I've commented in 1a80ce01a25a57489ef789f9d2609b84ecebf633

docker-compose-examples/keycloak-mysql.yml
stianst: I would remove it here. This example is meant as a starting point to use MySQL and if there is no specific reason to add it then it may confuse people.

Perhaps instead add a separate yml file to show custom JDBC_PARAMS would be better? Something like keycloak-mysql-custom-params.yml?

server/docker-entrypoint.sh
stianst: Thanks. My sed skills are a bit rusty, so couldn't quite figure out what it did ;)

docker-compose-examples/keycloak-mysql.yml
thaleshonda: I thought it would have many duplicated information in a keycloak-mysql-custom-params.yml file. Instead I commented the JDBC_PARAMS variable with a warning message in 15799ebe527e93dfc3ab4df2b5290440048cd110
What do you think?

stianst:
Looks good, thanks

PR #107, opened 2018-04-07
Using latest postgres JDBC 4.2 driver

matzew:
In March 2018 the Postgres team did release the `42.2.2` as its current version


stianst:
Thanks @matzew

PR #108, opened 2018-04-09
Generic DB scripts

ruromero:
I think there's no benefit in having Environment variables per each DB type. Having a generic name would allow Keycloak to connect to any of the 3 database types just by changing the variables value.

I haven't updated the documentation or templates yet. I wanted some feedback before continuing.


stianst:
This is a better approach, but we would still have to support the old variables as well. The reason for the different variables is a bit historical as there used to be separate images for each db. When we added support to all DBs to the main image we kept the variable names the same for backwards compatibility.

ruromero:
@stianst cool, that is what I wanted to know. I'll continue working on this during this week. Thanks for the feedback

ruromero:
@stianst can you review it? It should work with the specific variables as well.

stianst:
Before reviewing/testing it there's at least two things that should be updated:

* DB_VENDOR should be made mandatory now. With the DB specific variables deprecated we can't select the DB vendor based on the variables.
* MySQL and Postgres define their own variables. When starting the actual DB we have to use the values defined by them. For instance MySQL uses MYSQL_USER not DB_USER. This will need changing in the readme and compose examples

williamabarca01:
Hi guys Is it possible to support Oracle DB?
Thanks!

stianst:
@williamabarca01 I don't think it is due to licenses issues. If you'd like to discuss this further please use our user mailing lists as this is not the correct place.

ruromero:
@stianst changes done and merged with the latest changes

PR #115, opened 2018-05-13
Bumps MySQL Driver version to 5.1.46 to support MySQL 8

cristhiank:
I also sent an email to the keycloak dev list to discuss this. https://lists.jboss.org/pipermail/keycloak-dev/2018-May/010815.html


cristhiank:
Yes, it's only an upgrade to support a new authentication mechanism in MySQL 8 that is enabled by default as you can see [here](https://mysqlserverteam.com/mysql-8-0-4-new-default-authentication-plugin-caching_sha2_password/)
It will allow migrating existent databases but the bug with new databases will exist until https://github.com/keycloak/keycloak/pull/5201 is fixed.

stianst:
https://issues.jboss.org/browse/KEYCLOAK-7500

stianst:
Thanks

PR #116, opened 2018-05-28
Keycloak HA Profile

coraxs:
In image keycloak-ha-postgres, keycloak binds private interface to 127.0.0.1, by default. So because jgroups uses private interface for communicating, nodes cannot see each other, and cannot make cluster. In this fix nodes are using container IP for private interface (0.0.0.0 is not valid ip for udp).  Also IP of private interface can be overridden in CMD in Dockerfile, if desired. Additional files (add-admin.sh and add-datasource.sh) are created for reducing boilerplate code because new docker entrypoint is created. Ports 55200 and 45688 are exposed.
This will automate clustering, and enable quick scaling of nodes in docker compose. Like in readme.md example.

### DIFF
--- a/server-ha/Dockerfile
+++ b/server-ha/Dockerfile
@@ -1,0 +1,10 @@
+FROM jboss/keycloak:latest
+
+ADD docker-entrypoint-ha.sh /opt/jboss/docker-entrypoint-ha.sh
+
+EXPOSE 55200
+EXPOSE 45688
+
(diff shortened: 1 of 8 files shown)


ray-matos-mpf:
Someone should really look at this PR, as this is exactly what i was looking for.

stianst:

stianst:

slaskawi:
@evaryont Yes, it should be there. BTW, I will be working this week on getting us a solution for HA in containers. Please stay tuned!

PR #120, opened 2018-06-14
Update to 4.0.0.Final

stianst:


masteinhauser:
I see that this was merged in leu of my prior PR https://github.com/jboss-dockerfiles/keycloak/pull/113, but this PR ***does not*** actually update the proxy to 4.0.0.Final
https://github.com/jboss-dockerfiles/keycloak/blob/master/proxy/Dockerfile

PR #126, opened 2018-06-26
An example of connecting Keycloak to the Oracle database.

cleankod:


stianst:
Thanks for the PR. Sorry for not responding earlier.

In it's current form this can't be merged as it's not following the way we are adding support for databases. We are adding support directly to the main server image rather than creating DB specific images. You can take a look at https://github.com/jboss-dockerfiles/keycloak/pull/130 as a reference on how it should be added. I've not looked at that PR in detail yet, but it looks like it's correct.

I'm going to close this PR, but would be great if you would open a new PR after you've adopted the approach we are taking.

denyshetzel:
There is a possibility to pass a full url as examples:

`jdbc:oracle:thin:@HOSTNAME:PORT/SERVICENAME`

or

`jdbc:oracle:thin:@(description=(address=(host=HOSTNAME)(protocol=tcp)(port=PORT))(connect_data=(service_name=SERVICENAME)(server=SHARED)))`


I inspected the code snippet and I don't see how to do it, and it doesn't work when using **service name** Oracle DB

`...
/subsystem=datasources/data-source=KeycloakDS: add(jndi-name=java:jboss/datasources/KeycloakDS,enabled=true,use-java-context=true,use-ccm=true, connection-url=jdbc:oracle:thin:@${env.DB_ADDR:oracle}:${env.DB_PORT:1521}:${env.DB_DATABASE:XE}${env.JDBC_PARAMS:}, driver-name=oracle)
...`

PR #129, opened 2018-07-05
https-listener also honor proxy-address-forwarding setting

loekalive:
Ensure correct ip addresses show up in logging when https is used between proxy and keycloak.


stianst:
Thanks. Added JIRA https://issues.jboss.org/browse/KEYCLOAK-7860 for this.

PR #114, opened 2018-05-08
Add support for mssql

jgorin86:
Adding opensource mssql driver and configuration files to support Mssql as DB_VENDOR

PR #130, opened 2018-07-16
Add Microsoft SQL Server support

RolandTa:
Add Microsoft SQL Server support based on changes from jgorin86/keycloak and additional feedback from pull-request #114


Tomschi:
I also need the Microsoft SQL Server support.

stianst:
Thanks for the PR. I've only had a chance to have a brief look so far and all looks very good. I especially like the fact that you caught the breaking changes I just made and fix it almost instantaneously ;)

I will try to take a proper look next week so we can get this merged.

sschu:
We are also interested in this as we are essentially doing the same thing in our custom image. As a side note, the current version is 7.0.0, see https://github.com/Microsoft/mssql-jdbc/releases. Not sure its necessary to take this, we are currently on 6.4.0 as well.

server/tools/cli/databases/mssql/change-database.cli
```diff
@@ -0,0 +1,9 @@
+/subsystem=datasources/data-source=KeycloakDS: remove()
+/subsystem=datasources/data-source=KeycloakDS: add(jndi-name=java:jboss/datasources/KeycloakDS,enabled=true,use-java-context=true,use-ccm=true, connection-url="jdbc:sqlserver://${env.DB_ADDR:mssql}:${env.DB_PORT:1433};databaseName=${env.DB_DATABASE:keycloak};integratedSecurity=false;user=${env.DB_USER:keycloak};password=${env.DB_PASSWORD:password};${env.JDBC_PARAMS:}", driver-name=sqlserver)
```
sschu: Wouldn't it make sense to also add "sendStringParametersAsUnicode=false"? We are using this and reduced the amount of deadlocks quite substantially on mass data imports, see Hyneks comment on https://issues.jboss.org/browse/KEYCLOAK-3210.

server/Dockerfile
```diff
@@ -4,6 +4,7 @@ ENV KEYCLOAK_VERSION 4.3.0.Final
 ENV JDBC_POSTGRES_VERSION 42.2.2
 ENV JDBC_MYSQL_VERSION 5.1.46
 ENV JDBC_MARIADB_VERSION 2.2.3
+ENV JDBC_MSSQL_VERSION 6.4.0
```
sschu: If I am not mistaken, downloading the JDBC driver needs to be added to build-keycloak.sh

RolandTa:
@stianst Pull Request includes now all recent changes and feedback from @sschu regarding mssql jdbc 7.0 / missing driver download and sendStringParamtersAsUnicode=false

stianst:
Tried this from steps in README and it doesn't work as the user and password are wrong. The db is also not created. Can someone update the README to include setting DB_USER/DB_PASSWORD and also get it to create the Keycloak database, or at least provide a docker exec command to do that prior to starting Keycloak?

Also, I need this squashed to a single commit, which is awkward to do since it has merge commits in there.

michaeljmcd:
@stianst - looking at the SQL Server side of the house, it looks like the reason that the DB_USER/DB_PASSWORD setting and database autocreation aren't present is that the base MS SQL Server Docker image doesn't support them yet:

https://github.com/Microsoft/mssql-docker/issues/2

stianst:
That's fine, we just need to make sure that the README covers those aspects and that the example on how to run it includes the steps needed to get it working. At this point I couldn't even with giving it a bit of a go get it running.

michaeljmcd:
Completely fair point. Attached is a Docker Compose file I used to get it up and running based on some of the POC work I was doing. There is an additional image - `mssqlscripts` - that will attempt to create the Keycloak database every five seconds until it succeeds (it terminates after success).

[docker-compose.yml.txt](https://github.com/jboss-dockerfiles/keycloak/files/2458207/docker-compose.yml.txt)

abstractj:
@RolandTa @michaeljmcd I'm closing this PR for now. But feel free to reopen or open a new one, when you think it's ready to go.

PR #140, opened 2018-08-20
KEYCLOAK-8072 Set permissions in format expected for OpenShift

stianst:


slaskawi:
Good news! This PR shrinks the image by 20 MB:
```
REPOSITORY          TAG                 IMAGE ID            CREATED             SIZE
keycloak-dev        latest              4b77f9890149        13 minutes ago      780 MB
docker.io/jboss/keycloak   4.3.0.Final         75c249eec186        5 days ago          801 MB
```

It also seems to run correctly (haven't checked all use cases by it boots up and I can actually use it to secure my app).

PR #141, opened 2018-08-20
KEYCLOAK-8103 Remove deprecated images

stianst:


server-ha-postgres/README.md
```diff
@@ -1,5 +1,7 @@
 # Keycloak HA PostgreSQL
 
+This image is deprecated.
```
pmverma: Can you please describe what other options are available if this image is deprecated?
Thanks in advance.

stianst: The main server image already has support for Postgres and will soon also have support for clustering/HA, so there will be no need for this image anymore.

uniquestring: What does "soon" mean?
The latest ha image `jboss/keycloak-ha-postgres:4.3.0.Final` seems to be broken (using `latest` tag is not an option).

Should I wait for the `jboss/keycloak` image to officially support ha or should I just try to get the ha image working by editing the `standalone-ha.xml`?

Btw. I think it's a bit strange to remove the `jboss/keycloak-postgres` image from the DockerHub before `jboss/keycloak-ha-postgres`, because now the ha image is missing it's direct dependency (`FROM jboss/keycloak-postgres:4.3.0.Final `).

stianst: We're waiting for Keycloak 4.3 release as it will include upgrade to WildFly 13 which has significant improvements around ping mechanisms simplifying HA setup on OpenShift/Kubernetes/etc. 

The aim is to have HA support in the main image for Keycloak 4.4 (4-5 weeks from now).

I forgot that this image hadn't been updated to use jboss/keycloak yet. Should be fairly simply to change that.

If anyone wants to send a PR to fix this image that'd be great. Otherwise we'll have to wait for HA in jboss/keycloak.

uniquestring: Thank you for the info. Simplifying the HA setup sound promising.

As far as I understand it, you need a separate JGroups container when building a Keycloak cluster, because udp multicast isn't supported for some environments.

I tried creating a cluster with a spearate JGroups instance, but JGroups is throwing all kinds of errors as soon as the first container connects.
Will this be part of the planned simplification?

Btw.
I tried, but couldn't reproduce the problem I had with `4.3.0.Final`. I guess it had todo with the separate JGroups container, without it everything seems to be fine.

PR #143, opened 2018-08-22
KEYCLOAK-7891 OpenShift managed SSL certificates

slaskawi:
https://issues.jboss.org/browse/KEYCLOAK-7891

This PR introduces automatic generation of Keystore/Truststore based on provided PEM files. It has been implemented very [similarly to the product](https://github.com/jboss-openshift/cct_module/blob/master/os-sso72/added/launch/openshift-x509.sh) (mostly for symmetry reasons). However, there are some things that has been changed:

* I'm using CLI instead of manipulating XML configuration directly
* OpenSSL generated a warning message when generating new random values. I silenced it out.
* I removed extracting JGroups secret (I think it doesn't bring any value to the table at the moment).

Possible things to consider:

* CA Bundle path is specified by an environmental variable whereas `tls.crt` and `tls.key` are expected to be mounted in some path. This lack of symmetry can be fixed by either introducing 2 new environmental variables pointing to `tls.crt` and `tls.key`. Removing `X509_CA_BUNDLE` is probably not an option since CA bundle is in different place in OpenShift and Kubernetes.
* Introduce a switch to ignore Keystore/Truststore settings, set it to `true` (meaning, yes please ignore those settings) and squash `keycloak-https-mutual-tls.json` and `keycloak-https.json` together.

### DIFF
--- a/openshift-examples/README.md
+++ b/openshift-examples/README.md
@@ -1,0 +1,6 @@
+# Keycloak OpenShift examples
+
+This directory contains a set of predefined OpenShift templates for running Keycloak, including:
+
+* `keycloak-https.json` - A standard template for most of the use cases. It uses both HTTP and HTTPS routes.
+* `keycloak-https-mutual-tls.json` - A similar template to the one above but uses OpenShift generated certificates to setup Mutual TLS.
--- a/openshift-examples/keycloak-https-mutual-tls.json
+++ b/openshift-examples/keycloak-https-mutual-tls.json
@@ -1,0 +1,242 @@
+{
+    "kind": "Template",
+    "apiVersion": "v1",
+    "metadata": {
+        "name": "keycloak-https",
+        "annotations": {
+            "iconClass": "icon-sso",
+            "tags": "keycloak",
+            "version": "4.0.0.Beta2",
+            "openshift.io/display-name": "Keycloak",
+            "description": "An example Keycloak server with HTTPS"
+        }
+    },
+    "parameters": [
+        {
+            "displayName": "Application Name",
+            "description": "The name for the application.",
+            "name": "APPLICATION_NAME",
+            "value": "keycloak",
+            "required": true
+        },
+        {
+            "displayName": "Keycloak Administrator Username",
+            "description": "Keycloak Server administrator username",
+            "name": "KEYCLOAK_USER",
+            "from": "[a-zA-Z0-9]{8}",
+            "generate": "expression",
+            "required": true
+        },
+        {
+            "displayName": "Keycloak Administrator Password",
+            "description": "Keycloak Server administrator password",
+            "name": "KEYCLOAK_PASSWORD",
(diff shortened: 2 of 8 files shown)


slaskawi:
This can be tested by creating a pod in OpenShift and calling:

```
openssl s_client -key ./tls.key -cert ./tls.crt -CAfile /var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt -connect secure-keycloak.myproject.svc:8443 -verify_return_error -verify 10 -msg
```

the output should be somewhat like this:

```
SSL-Session:
    Protocol  : TLSv1.2
    Cipher    : ECDHE-RSA-AES256-SHA384
    Session-ID: 5B7E8746942A53C7B1F0FE9285C349B100F518E7D2F73B9653054764FE2040EA
    Session-ID-ctx: 
    Master-Key: 843428CBBF526041C798EBB95E47A9CB33C7C242C20212BF0BB5CA1587F38C2DB588BF628FB923C643FBF024BFC0FF17
    PSK identity: None
    PSK identity hint: None
    SRP username: None
    Start Time: 1535018822
    Timeout   : 7200 (sec)
    Verify return code: 0 (ok)
    Extended master secret: yes
```

dash042:
trying to get this to work on a Docker Swarm with own certificates and ca-bundle and testing (as described in readme with curl) will throw:
> {"error_description":"X509 client certificate is missing.","error":"invalid_request"}

Question: is adding the SSL-REALM and the https-listener/default (which already exists in the new 4.4.0.Final) - see https://www.keycloak.org/docs/latest/server_admin/index.html#enable-x-509-client-certificate-user-authentication still required?

slaskawi:
@dash042 You need to alter the Elytron subsystem. See [the manual](https://docs.jboss.org/author/display/WFLY/SSL+Configuration+using+Elytron+Subsystem). You may also have a look at my sample configuration in [here](https://gist.github.com/slaskawi/92aead6c519b867621129b640b4a3c88#file-standalone-xml-L311-L335) and [here](https://gist.github.com/slaskawi/92aead6c519b867621129b640b4a3c88#file-standalone-xml-L573).

dash042:
@slaskawi running keycloak as suggested (Docker) already makes this/your configuration:
`docker run -d -p 8080:8080 -p 8443:8443 --name keycloak \
-v /root/sso/server:/etc/x509/https -v /root/ca/intermediate/certs/ca-chain.cert.pem:/ca-chain.cert.pem \
-e KEYCLOAK_USER=admin -e KEYCLOAK_PASSWORD=test123 -e X509_CA_BUNDLE=/ca-chain.cert.pem \
jboss/keycloak:4.4.0.Final`

so the SSL-REALM settings are not required?

slaskawi:
Got you now. No, you don't need SSL-REALM settings. Everything is done through Elytron.

BTW, you need to use an absolute path in `X509_CA_BUNDLE`.

PR #144, opened 2018-08-27
Feature/fix docker entrypoint

jntesteves:
Keycloak fails to start with a DB, always falls back to H2

The second commit fixes the following message during startup:

> The batch failed with the following error: : File /opt/jboss/keycloak/cli/databases/mysql/change-database.cli does not exist.

I couldn't fix the root cause because I don't know what is trying to access this path, hence the simple workaround with a symlink.

### DIFF
--- a/server/tools/docker-entrypoint.sh
+++ b/server/tools/docker-entrypoint.sh
@@ -91,6 +91,7 @@ echo "========================================================================="
 echo ""
 
 if [ "$DB_VENDOR" != "h2" ]; then
+    ln -s /opt/jboss/tools/cli /opt/jboss/keycloak/cli
     /bin/sh /opt/jboss/tools/databases/change-database.sh $DB_VENDOR
 fi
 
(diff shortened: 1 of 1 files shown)


stianst:
What are the steps to reproduce the issue? I can't accept this workaround.

jntesteves:
Hello, I'm just building Keycloak 4.3.0.Final from my repo as explained here https://github.com/jboss-dockerfiles/keycloak/tree/master/server#building-image-with-keycloak-from-different-sources
When I try to start my compose with MySQL I get that message in the logs twice and fallback to H2. Nothing else interesting gets logged.

I did this workaround mainly to get back on track with my development, I suppose it must be fixed somewhere else in the future, but I thought I would post a PR anyway just in case people need it right now and no one is working in the root cause.

The fix from the first commit is safe to apply, it's just a wrong path. Don't need to be applied from this PR.

stianst:
Found the problems and fixed them here https://github.com/jboss-dockerfiles/keycloak/pull/145. So closing this PR. Thanks

PR #146, opened 2018-08-30
Add environment variables for fixed hostname http and https ports

yoransabern:


stianst:
Thanks for your PR. I don't think this will be needed anymore as I made some improvements on how the ports are handled. See https://issues.jboss.org/browse/KEYCLOAK-8069. As long as any proxies are configured properly the port will automatically be retrieved from the request so there is no need to set this manually.

yoransabern:
Hi, the problem is that we can't change the request ports in our system, internal calls go over 8080, but we don't want to append these ports for links in emails etc., having the abillity to set these ports in the docker image solves this issue for us.

E-2:
Hi,
In our use case (colleague of yoransabern here) this is the only thing missing from calling it internally between docker services.
The workaround for us at the moment is to use the external url, but this routes it unnecessary through AWS ELB

stianst:
Ah of course. If you have clients calling internal URL/IP with a different port to the external URL you will still need to fix it. Will review and merge this soon.

stianst:
Thanks

PR #149, opened 2018-09-10
Downloading and installation the custom themes based on $CUSTOM_THEMES_DIST env variable

wildloop:
What about this idea?
This is a proven way that users can use to customize themes without modifying container.
What about doing plugins, providers, etc it the same way? 
(Please correct my grammar mistakes in readme.md)

### DIFF
--- a/server/README.md
+++ b/server/README.md
@@ -144,6 +144,7 @@ found here:
 ## Adding custom theme
 
 To add a custom theme extend the Keycloak image add the theme to the `/opt/jboss/keycloak/themes` directory.
+You can also specify the environment variable `CUSTOM_THEMES_DIST` with url to tar.gz package containing your custom themes folders. The package will be downloaded and installed on container start.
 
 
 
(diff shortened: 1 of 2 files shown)


stianst:
Thanks for your PR. 

This is not a great approach to adding themes and providers. It requires hosting the bits somewhere, then it will result in downloading this file every time a container is started. In Kubernetes/OpenShift where pods are spun up regularly that's not a good thing.

For development it would be better to hot-deploy to the server, or mount a local folder.

For production it's better to extend the image to add the bits, or to use something like OpenShift s2i to have the "extended" images built for you.

With that in mind I'm going to reject this PR as it's not something we would like to have in the image.

wildloop:
OK thanks. 
However, it would be nice to find a way to add themes without extending the image.

stianst:
I honestly can't see that happening. The option to dl at runtime is not a good option, which only leaves extending the image directly or using something that automatically builds images for you (there's loads of things that do that).

stianst:
Actually you have other options that don't require extending the image. Mounting a volume, using a config map (if you're in Kubernetes or OpenShift), etc..

wildloop:
I think that it's not worst idea to download themes in the release stage, that takes the build produced by the build stage and combines it with the deploy’s current config (in this case: custom themes) - https://12factor.net/build-release-run

Mounting the volume with the themes (nfs as a service) makes app a little bit depended on env - rule: https://12factor.net/dependencies

So, you are right - the best I can do is to form and customize image :(

PR #151, opened 2018-09-18
KEYCLOAK-7582 Discovery protocol support

slaskawi:
https://issues.jboss.org/browse/KEYCLOAK-7582

This PR adds clustering support for the Keycloak Docker image. As the matter of fact, it makes allows to use any discovery protocol with custom configuration.

Here's how it works:
* The configuration scripts pick up `JGROUPS_DISCOVERY_PROTOCOL` and `JGROUPS_DISCOVERY_PROPERTIES` variables.
* In the next step, the main entry point script passes those two configuration variables to `jgroups.sh` script.
* The script adjusts the variables and invokes `cli/jgroups/discovery/default.cli` (or a cli file with an exact protocol name if it exists; but in most of the cases, the `default.cli` should be enough).
* The CLI script modifies `standalone-ha.xml`. It removed the default discovery protocol and replaces it with the one specified above.
* The final piece of the puzzle is adjusting JGroups bind options. By default they operate on the `private` interface. We need rewire it to the external, public port. By default we use `hostname -i` for that.

I also uploaded a test image, so that you can play with it. Just replace the `image` tag in the template with `slaskawi/keycloak-jgroups-discovery`.

### DIFF
--- a/openshift-examples/keycloak-https-mutual-tls.json
+++ b/openshift-examples/keycloak-https-mutual-tls.json
@@ -55,6 +55,12 @@
             "name": "HOSTNAME_HTTPS",
             "value": "",
             "required": false
+        },
+        {
+            "displayName": "Namespace used for DNS discovery",
+            "description": "This namespace is a part of DNS query sent to Kubernetes API. This query allows the DNS_PING protocol to extract cluster members. This parameter might be removed once https://issues.jboss.org/browse/JGRP-2292 is implemented.",
+            "name": "NAMESPACE",
+            "required": true
         }
     ],
     "objects": [
@@ -212,6 +218,14 @@
                                     {
                                         "name": "X509_CA_BUNDLE",
                                         "value": "/var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt"
+                                    },
+                                    {
+                                        "name": "JGROUPS_DISCOVERY_PROTOCOL",
+                                        "value": "dns.DNS_PING"
+                                    },
+                                    {
+                                        "name": "JGROUPS_DISCOVERY_PROPERTIES",
+                                        "value": "${APPLICATION_NAME}.${NAMESPACE}.svc.cluster.local"
                                     }
                                 ],
                                 "securityContext": {
--- a/openshift-examples/keycloak-https.json
+++ b/openshift-examples/keycloak-https.json
@@ -55,6 +55,12 @@
             "name": "HOSTNAME_HTTPS",
             "value": "",
             "required": false
+        },
+        {
+            "displayName": "Namespace used for DNS discovery",
+            "description": "This namespace is a part of DNS query sent to Kubernetes API. This query allows the DNS_PING protocol to extract cluster members. This parameter might be removed once https://issues.jboss.org/browse/JGRP-2292 is implemented.",
+            "name": "NAMESPACE",
+            "required": true
         }
     ],
     "objects": [
@@ -207,6 +213,14 @@
                                     {
                                         "name": "DB_VENDOR",
                                         "value": "${DB_VENDOR}"
+                                    },
+                                    {
+                                        "name": "JGROUPS_DISCOVERY_PROTOCOL",
+                                        "value": "dns.DNS_PING"
+                                    },
+                                    {
+                                        "name": "JGROUPS_DISCOVERY_PROPERTIES",
+                                        "value": "${APPLICATION_NAME}.${NAMESPACE}.svc.cluster.local"
                                     }
                                 ],
                                 "securityContext": {
--- a/server/README.md
+++ b/server/README.md
@@ -141,17 +141,75 @@ found here:
 
 
 
-## Adding custom theme
+## Adding a custom theme
 
 To add a custom theme extend the Keycloak image add the theme to the `/opt/jboss/keycloak/themes` directory.
 
 
 
-## Adding custom provider
+## Adding a custom provider
 
 To add a custom provider extend the Keycloak image and add the provider to the `/opt/jboss/keycloak/standalone/deployments/`
 directory.
 
+
+## Clustering
+
+Replacing the default discovery protocols (`PING` for the UDP stack and `MPING` for the TCP one) can be achieved by defining
+two additional environment variables:
+
+- `JGROUPS_DISCOVERY_PROTOCOL` - name of the discovery protocol, e.g. DNS_PING
+- `JGROUPS_DISCOVERY_PROPERTIES` - an optional parameter with the discovery protocol properties in the following format:
+`PROP1=FOO,PROP2=BAR`
+
+The bootstrap script will detect the variables and adjust the `standalone-ha.xml` configuration based on them.
+
+### PING example
+
+The `PING` discovery protocol is used by default in `udp` stack (which is used by default in `standalone-ha.xml`).
+Since the Keycloak image runs in clustered mode by default, all you need to do is to run it:
+
+    docker run jboss/keycloak
+
+If you two instances of it locally, you will notice that they form a cluster.
+
+### OpenShift example with dns.DNS_PING
+
+Clustering for OpenShift can be achieved by using either `dns.DNS_PING` and a governing service or `kubernetes.KUBE_PING`.
+The latter requires `view` permissions which are not granted by default, so we suggest using `dns.DNS_PING`.
+
+#### Using the template
+
+The full example has been put into the `openshift-examples` directory. Just run one of the two templates. Here's an example:
+
+    oc new-app -p NAMESPACE=`oc project -q` -f keycloak-https.json
+
+#### What happen under the hood?
+
+Both OpenShift templates use `dns.DNS_PING` under the hood. Here's an equivalent docker-based command that OpenShift
+is invoking:
+
+    docker run \
+    -e JGROUPS_DISCOVERY_PROTOCOL=dns.DNS_PING -e \
+    JGROUPS_DISCOVERY_PROPERTIES=dns_query=keycloak.myproject.svc.cluster.local \
+    jboss/keycloak
+
+In this example the `dns.DNS_PING` that queries `A` records from the DNS Server with the following query
+`keycloak.myproject.svc.cluster.local`.
+
+### Adding custom discovery protocols
+
+The default mechanism for adding discovery protocols should cover most of the cases. However, sometimes
+you need to add more protocols at the same time, or adjust other protocols. In such cases you will need
+your own cli file placed in `/opt/jboss/tools/cli/jgroups/discovery`. The `JGROUPS_DISCOVERY_PROTOCOL` need to
+match your cli file, for example:
+
+    JGROUPS_DISCOVERY_PROTOCOL=custom_protocol
+    /opt/jboss/tools/cli/jgroups/discovery/custom_protocol.cli
+
+This can be easily achieved by extending the Keycloak image and adding just one file.
+
+Of course, we highly encourage you to contribute your custom scripts back to the community image!
 
 
 ## Misc
--- a/server/tools/cli/jgroups/discovery/default.cli
+++ b/server/tools/cli/jgroups/discovery/default.cli
@@ -1,0 +1,9 @@
+embed-server --server-config=standalone-ha.xml --std-out=echo
+batch
+/subsystem=jgroups/stack=udp/protocol=PING:remove()
+/subsystem=jgroups/stack=udp/protocol=$keycloak_jgroups_discovery_protocol:add(add-index=0, properties=$keycloak_jgroups_discovery_protocol_properties)
+
+/subsystem=jgroups/stack=tcp/protocol=MPING:remove()
+/subsystem=jgroups/stack=tcp/protocol=$keycloak_jgroups_discovery_protocol:add(add-index=0, properties=$keycloak_jgroups_discovery_protocol_properties)
+run-batch
+stop-embedded-server
--- a/server/tools/docker-entrypoint.sh
+++ b/server/tools/docker-entrypoint.sh
@@ -24,6 +24,27 @@ if [ "$KEYCLOAK_HOSTNAME" != "" ]; then
     fi
 fi
 
+########################
+# JGroups bind options #
+########################
+
+if [ -z "$BIND" ]; then
+    BIND=$(hostname -i)
+fi
+if [ -z "$BIND_OPTS" ]; then
+    BIND_OPTS="-Djboss.bind.address=$BIND -Djboss.bind.address.private=$BIND"
+fi
+SYS_PROPS+=" $BIND_OPTS"
+
+#################
+# Configuration #
+#################
+
+# If the "-c" parameter is not present, append the HA profile.
+if echo "$@" | egrep -v -- "-c "; then
+    SYS_PROPS+=" -c standalone-ha.xml"
+fi
+
 ############
 # DB setup #
 ############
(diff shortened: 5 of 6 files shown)


openshift-examples/keycloak-https-mutual-tls.json
```diff
@@ -184,7 +191,8 @@
                         "containers": [
                             {
                                 "name": "${APPLICATION_NAME}",
-                                "image": "jboss/keycloak-openshift",
+                                "image": "slaskawi/keycloak-jgroups-discovery",
```
pedroigor: Do we need to change the image here?

pedroigor:
LGTM

openshift-examples/keycloak-https-mutual-tls.json
slaskawi: No! definitely not.... that one slipped in by an accident... fixing.

server/tools/docker-entrypoint.sh
```diff
@@ -24,6 +24,18 @@ if [ "$KEYCLOAK_HOSTNAME" != "" ]; then
...
 
+########################
+# JGroups bind options #
+########################
+
+if [ -z "$BIND" ]; then
+    BIND=$(hostname -i)
+fi
+if [ -z "$BIND_OPTS" ]; then
+    BIND_OPTS="-Djboss.bind.address.management=0.0.0.0 -Djgroups.join_timeout=1000 -Djgroups.bind_addr=$BIND -Djboss.bind.address=$BIND"
```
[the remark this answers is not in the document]
rhusar: Also, what's the reasoning behind GMS.join_timeout of 1 second? (Seems quite low to me for a cloud environment and I was thinking we may actually increase it from the default number for cloud images)

deitch:
I am coming to this from #96. Once merged, what would be the "official" way of running this in Kubernetes? Would we just use `DNS_PING` and set up a headless service with `ClusterIP=None`, thus resolving all of the affiliated IPs?

stianst:
Haven't reviewed this yet, but realized we need a check to make sure clustering isn't enabled with embedded H2 dB as those two together makes no sense

server/tools/docker-entrypoint.sh
rhusar: The `jgroups.bind_addr`  has effect in WildFly-based software using jgroups subsystem, which I assume is the case here.

openshift-examples/keycloak-https-mutual-tls.json
```diff
@@ -55,6 +55,13 @@
             "name": "HOSTNAME_HTTPS",
             "value": "",
             "required": false
+        },
+        {
+            "displayName": "Namespace used for DNS discovery",
+            "description": "This namespace is a part of DNS query sent to Kubernetes API. This query allows the DNS_PING protocol to extract cluster members. This parameter might be removed once https://issues.jboss.org/browse/JGRP-2292 is implemented.",
+            "name": "NAMESPACE",
+            "value": "myproject",
```
stianst: Can the namespace somehow be templated so this doesn't have to be changed? Surely OpenShift knows what namespace it's being deployed to?

openshift-examples/keycloak-https-mutual-tls.json
```diff
@@ -185,6 +192,7 @@
                             {
                                 "name": "${APPLICATION_NAME}",
                                 "image": "jboss/keycloak-openshift",
+                                "args": ["-c", "standalone-ha.xml"],
```
stianst: Shouldn't changing to standalone-ha.xml be handled by the image itself rather than having to change the args here?

openshift-examples/keycloak-https-mutual-tls.json
slaskawi: Unfortunately no. And as the parameter description mentions, JGRP-2292 prevents us from using a Service name (since it appends `svc.cluster.local` suffix). See the [JGRP-2292](https://issues.jboss.org/browse/JGRP-2292) ticket for details. 

It isn't perfect, I know. But we'll improve it once the JGroups patch is in.

server/README.md
```diff
@@ -141,18 +141,33 @@ found here:
...
+two additional environment variables:
+
+* `JGROUPS_DISCOVERY_PROTOCOL` - name of the discovery protocol, e.g. DNS_PING
+* `$JGROUPS_DISCOVERY_PROPERTIES` - an optional parameter with the discovery protocol properties in the following format:
+`PROP1=FOO,PROP2=BAR`
+
+Here's an example that adds `DNS_PING` that queries `A` records from DNS with the following query `keycloak.myproject.svc.cluster.local`:
+
+    docker run \
+    -e JGROUPS_DISCOVERY_PROTOCOL=DNS_PING -e \
```
petr-k: Shouldn't the protocol name be `dns.DNS_PING`?

server/tools/docker-entrypoint.sh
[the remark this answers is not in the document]
slaskawi: Sorry, I just copied it from Infinispan template without putting much thought into it. Let me adjust it to our needs. Thanks a lot for spotting this Rado!

server/tools/jgroups.sh
```diff
@@ -0,0 +1,22 @@
...
+    # If there's a specific CLI file for given protocol - execute it. If not, we should be good with the default one.
+    if [ -f "/opt/jboss/tools/cli/jgroups/discovery/$JGROUPS_DISCOVERY_PROTOCOL.cli" ]; then
+       $JBOSS_HOME/bin/jboss-cli.sh --file="/opt/jboss/tools/cli/jgroups/discovery/$JGROUPS_DISCOVERY_PROTOCOL.cli" >& /dev/null
+    else
+       $JBOSS_HOME/bin/jboss-cli.sh --file="/opt/jboss/tools/cli/jgroups/discovery/default.cli" >& /dev/null
+    fi
+fi
+
+
+
```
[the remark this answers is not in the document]
slaskawi: Done.

slaskawi:
> I am coming to this from #96. Once merged, what would be the "official" way of running this in Kubernetes? Would we just use DNS_PING and set up a headless service with ClusterIP=None, thus resolving all of the affiliated IPs?

@deitch The `DNS_PING` is the suggested way of doing that. The biggest pain of `KUBE_PING` is that it needs a `view` permissions, which are not granted by default in OpenShift. Apart from that, feel free to use either `DNS_PING` or `KUBE_PING` (or even `PING/MPING` if your environment supports IP Multicasting; this feature needs to be explicitly turned on in the OpenShift project if that's the environment you're using).

server/README.md
slaskawi: Good point! Thanks!

openshift-examples/keycloak-https-mutual-tls.json
slaskawi: Ok, so I had a pretty long chat with Infinispan folks. In short - the replicated caches (used in `standalone-ha.xml`) will start slower (since they try to do discovery) but the runtime performance should be roughly the same as local ones. 

So it makes sense to always start the HA profile. I will add necessary switches.

slaskawi: Done.

slaskawi:
> Haven't reviewed this yet, but realized we need a check to make sure clustering isn't enabled with embedded H2 dB as those two together makes no sense

This one obviously requires a bit more work. I created https://issues.jboss.org/browse/KEYCLOAK-8331 to track this.

deitch:
@slaskawi wrote:

> The DNS_PING is the suggested way of doing that. The biggest pain of KUBE_PING is that it needs a view permissions, which are not granted by default in OpenShift.

I think this is great. I don't want to add additional permissions/roles to the service account under which keycloak runs - which, in any case, is an app, not a kube system-level service, so shouldn't get special kube rights to run - but more importantly it makes it generic. If I change my DNS, or service mesh, or whatever, DNS will "just work".

Thanks @slaskawi

server/tools/cli/jgroups/discovery/default.cli
```diff
@@ -0,0 +1,9 @@
+embed-server --server-config=standalone-ha.xml --std-out=echo
+batch
+/subsystem=jgroups/stack=udp/protocol=PING:remove()
+/subsystem=jgroups/stack=udp/protocol=$keycloak_jgroups_discovery_protocol:add(add-index=0, properties=$keycloak_jgroups_discovery_protocol_properties)
+
+/subsystem=jgroups/stack=tcp/protocol=MPING:remove()
```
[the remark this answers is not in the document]
slaskawi: I think we confuse things here... `TCP` or `UDP` is transport. `KUBE_PING`, `PING`, `MPING` or `dns.DNS_PING` are discovery protocols. You can not use more than one transport at a time (even if you create a fork channel). So I think there is no need to allow setting the discovery protocol separately per individual stack. 

Does this answer your question?

slaskawi:
@stianst 
> OK, as it's very likely that myproject won't be the right namespace I would leave it unspecified so users are required to set a value. As it is now there's a chance that users forget to set it and it's left as the default myproject value

This parameter will go away very shortly. We need two JGroups tickets to fix it ([JGRP-2295](https://issues.jboss.org/browse/JGRP-2295) and [JGRP-2292](https://issues.jboss.org/browse/JGRP-2292)). Rob already sent a pull request for both. So if we are lucky, this parameter will disappear in a couple of weeks.

So what do you think, having the above in mind, should we still enforce this parameter from users? Even though it will be gone shortly?

server/tools/jgroups.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
slaskawi: I don't agree here. 

I can imagine use cases where you will need multiple protocols for discovery (as we discussed a few days ago). Or maybe you will need to adjust `FD_*` protocol parameters with discovery timeouts (this is often the case when using `KUBE_PING` that polls Kubernetes API in a loop). In such case you will need to provide your own cli file that does it. As it is, it's super easy, just extend our image and drop a file into `/opt/jboss/tools/cli/jgroups/discovery`. If we removed this mechanism, you would need to drop a cli file and then copy-paste the `jgroups.sh` file. Now if you copy that file, you will need to maintain it manually by syncing manually changes we made in it with your own copy. That's a maintenance pain that is very easy to avoid with my proposed approach.

However, you might have a good point that this behavior should be described in the README. I'll add it.

openshift-examples/keycloak-https-mutual-tls.json
slaskawi: This parameter will go away very shortly. We need two JGroups tickets to fix it (JGRP-2295 and JGRP-2292). Rob already sent a pull request for both. So if we are lucky, this parameter will disappear in a couple of weeks.

So what do you think, having the above in mind, should we still enforce this parameter from users? Even though it will be gone shortly?

slaskawi: After giving it a little more thought... we can always remove that parameter once all changes are in. Give me a sec to introduce a fix.

server/tools/cli/jgroups/discovery/default.cli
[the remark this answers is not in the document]
petr-k: I'm not referring to the TCP/UDP protocols, but to JGroups named stacks. If you inspect the `standalone-ha.xml` file, there are two stacks named `tcp` and `udp` - and they use `TCP` and `UDP` protocols, respectively. JGroups defaults to the `udp` stack. At least that's my understanding. Or does it handle TCP and UDP at the same time?

My line of thought is this: if the configuration results in a UDP-based communication, then in Kubernetes you'd need to have a headless `Service` over `UDP` as well, which is somewhat unexpected and is not the default.

server/README.md
```diff
@@ -141,18 +141,33 @@ found here:
...
+two additional environment variables:
+
+* `JGROUPS_DISCOVERY_PROTOCOL` - name of the discovery protocol, e.g. DNS_PING
+* `$JGROUPS_DISCOVERY_PROPERTIES` - an optional parameter with the discovery protocol properties in the following format:
+`PROP1=FOO,PROP2=BAR`
+
+Here's an example that adds `DNS_PING` that queries `A` records from DNS with the following query `keycloak.myproject.svc.cluster.local`:
+
+    docker run \
+    -e JGROUPS_DISCOVERY_PROTOCOL=dns.DNS_PING -e \
```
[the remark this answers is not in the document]
slaskawi: Added proper manual entries. I hope it will look good now.

server/tools/jgroups.sh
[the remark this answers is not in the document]
slaskawi: Added necessary manual entries.

slaskawi:
@deitch 
> I think this is great. I don't want to add additional permissions/roles to the service account under which keycloak runs - which, in any case, is an app, not a kube system-level service, so shouldn't get special kube rights to run - but more importantly it makes it generic. If I change my DNS, or service mesh, or whatever, DNS will "just work".

You're welcome! Just remember about creating a governing service and/or specifying proper namespace when running the template.

server/tools/cli/jgroups/discovery/default.cli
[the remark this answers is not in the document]
petr-k: Nevermind. Since the headless service is used for DNS discovery only, it should not matter what protocol is used.

openshift-examples/keycloak-https-mutual-tls.json
```diff
@@ -212,6 +218,14 @@
...
                                         "name": "X509_CA_BUNDLE",
                                         "value": "/var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt"
+                                    },
+                                    {
+                                        "name": "JGROUPS_DISCOVERY_PROTOCOL",
+                                        "value": "dns.DNS_PING"
+                                    },
+                                    {
+                                        "name": "JGROUPS_DISCOVERY_PROPERTIES",
+                                        "value": "${APPLICATION_NAME}.${NAMESPACE}.svc.cluster.local"
```
petr-k: Shouldn't the `value` be `dns_query=${APPLICATION_NAME}.${NAMESPACE}.svc.cluster.local`?

deitch:
Really nice to see this merged in, thank you!

What release will this hit? And when will that be available at https://hub.docker.com/r/jboss/keycloak/

openshift-examples/keycloak-https-mutual-tls.json
slaskawi: At the moment, no. The `svc.cluster.local` is appended automatically. However, we consider this as a mistake and there's already a [JIRA](https://issues.jboss.org/browse/JGRP-2295) to remove it.

openshift-examples/keycloak-https.json
```diff
@@ -207,6 +213,14 @@
...
                                         "name": "DB_VENDOR",
                                         "value": "${DB_VENDOR}"
+                                    },
+                                    {
+                                        "name": "JGROUPS_DISCOVERY_PROTOCOL",
+                                        "value": "dns.DNS_PING"
+                                    },
+                                    {
+                                        "name": "JGROUPS_DISCOVERY_PROPERTIES",
+                                        "value": "${APPLICATION_NAME}.${NAMESPACE}.svc.cluster.local"
```
rhusar: This seems wrong, it should be `"value": "dns_query=${APPLICATION_NAME}.${NAMESPACE}.svc.cluster.local"`

openshift-examples/keycloak-https-mutual-tls.json
rhusar: This seems wrong, it should be `"value": "dns_query=${APPLICATION_NAME}.${NAMESPACE}.svc.cluster.local"`

PR #153, opened 2018-09-25
add possibility to change welcome theme by env variable

micedre:
This PR would permit to modify the welcome theme without the need to extends the docker image.
It simply update `standalone(-ha).xml` with an env value for `welcomeTheme` attribute (defaults to keycloak).


stianst:
Looks good, can you create a JIRA for this please?

micedre:
I have created https://issues.jboss.org/browse/KEYCLOAK-8416 , is it enough ?

stianst:
That's great. Thanks

PR #155, opened 2018-10-04
extended parametrization of environment variables and description of …

cjohn001:
Hello together, I tried to use the template with an external mysql database in Openshift and failed to deploy keycloak due to missing parameters, and because of two bugs in the keycloak init scripts. I am not fluent enough in bash scripting to get these bugs solved cleanly. However, the modified template now makes it more difficult for others to run into the same trap. As I do not know where the right place in this project is for posting of bug reports, I add them to the discussion here.

Bug 1: If using "mysql" as the name for the database pod, the init script seems to set a variable  MYSQL_PORT, with a wrong parametrization, which is than also placed into DB_PORT, see http://lists.jboss.org/pipermail/keycloak-user/2018-October/015737.html for a more detailed description.

Bug 2: In case a different name for the database pod is set, still the DB_PORT variable is set to some wrong value "tcp://ip:port"

Best regards,
 Christoph

### DIFF
--- a/openshift-examples/keycloak-https.json
+++ b/openshift-examples/keycloak-https.json
@@ -265,7 +265,7 @@
                                     },
                                     {
                                         "name": "DB_ADDR",
-                                        "value": "${DB_ADDR}.${NAMESPACE}.svc.cluster.local"
+                                        "value": "${DB_ADDR}"
                                     },
                                     {
                                         "name": "DB_PORT",
--- a/openshift-examples/keycloak-https.json
+++ b/openshift-examples/keycloak-https.json
@@ -8,7 +8,7 @@
             "tags": "keycloak",
             "version": "4.0.0.Beta2",
             "openshift.io/display-name": "Keycloak",
-            "description": "An example Keycloak server with HTTPS"
+            "description": "An example Keycloak server with HTTPS. \n\nNOTE: If you are going to use an external database, you need to manualy inject the environment variables DB_USER and DB_PASSWORD from your database secret."
         }
     },
     "parameters": [
@@ -16,6 +16,7 @@
             "displayName": "Application Name",
             "description": "The name for the application.",
             "name": "APPLICATION_NAME",
+            "from": "maximum length of 253 lower case alphanumeric characters, -, and .",
             "value": "keycloak",
             "required": true
         },
@@ -43,6 +44,41 @@
             "required": true
         },
         {
+            "displayName": "DB ADDR",
+            "description": "Specify the name of the database pod if you are going to use an external DB. The name of the DB pod should not be 'mysql' due to an keycloak internal bug",
+            "name": "DB_ADDR",
+            "from": "maximum length of 253 lower case alphanumeric characters, -, and .",
+            "value": "",
+            "required": false
+        },
+        {
+            "displayName": "DB Port",
+            "description": "DB port (default ports for H2 = 9123, MYSQL and MARIADB = 3306, POSTGRES = 5432",
+            "name": "DB_PORT",
+            "value": "9123",
+            "required": true
+        },
+        {
+            "displayName": "DB Database",
+            "description": "Name of the DB scheme",
+            "name": "DB_DATABASE",
+            "value": "keycloak",
+            "required": true
+        },
+        {
+            "displayName": "JDBC Parameters",
+            "description": "additional database driver parameters, if you use 'useSSL=true', you manually need to add a certificate",
+            "name": "JDBC_PARAMS",
+            "value": "useSSL=false",
+            "required": true
+        },
+        {
+            "displayName": "Namespace used for DNS discovery",
+            "description": "This namespace is a part of DNS query sent to Kubernetes API. This query allows the DNS_PING protocol to extract cluster members. This parameter might be removed once https://issues.jboss.org/browse/JGRP-2292 is implemented. Should be project name without '.svc.cluster.local' ",
+            "name": "NAMESPACE",
+            "required": true
+        },
+        {
             "displayName": "Custom http Route Hostname",
             "description": "Custom hostname for http service route. Leave blank for default hostname, e.g.: <application-name>.<project>.<default-domain-suffix>",
             "name": "HOSTNAME_HTTP",
@@ -55,12 +91,6 @@
             "name": "HOSTNAME_HTTPS",
             "value": "",
             "required": false
-        },
-        {
-            "displayName": "Namespace used for DNS discovery",
-            "description": "This namespace is a part of DNS query sent to Kubernetes API. This query allows the DNS_PING protocol to extract cluster members. This parameter might be removed once https://issues.jboss.org/browse/JGRP-2292 is implemented.",
-            "name": "NAMESPACE",
-            "required": true
         }
     ],
     "objects": [
@@ -240,7 +270,12 @@
                                     {
                                         "name": "JGROUPS_DISCOVERY_PROPERTIES",
                                         "value": "${APPLICATION_NAME}.${NAMESPACE}.svc.cluster.local"
+                                    },
+                                    {
+                                        "name": "DB_ADDR",
+                                        "value": "${DB_ADDR}.${NAMESPACE}.svc.cluster.local"
                                     }
+                                    
                                 ],
                                 "securityContext": {
                                     "privileged": false
--- a/openshift-examples/keycloak-https.json
+++ b/openshift-examples/keycloak-https.json
@@ -264,18 +264,29 @@
                                         "value": "${DB_VENDOR}"
                                     },
                                     {
+                                        "name": "DB_ADDR",
+                                        "value": "${DB_ADDR}.${NAMESPACE}.svc.cluster.local"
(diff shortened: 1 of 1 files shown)


abstractj:
@cjohn001 thank you, please file a Jira following the steps here https://www.keycloak.org/community.html. Make sure you add the steps to reproduce and attach your PR. In this way people will be able to review and add comments about it.

cjohn001:
ok, I will do that. Please pull the later changes as well. I had forgotten to commit the second half where the environment variables are actually set. Sorry, gets to late :)

stianst:
@cjohn001 not sure what's going on with the issues you're having. Would need more details and ideally the application template or steps you run on OpenShift to reproduce. Please make sure you attach that to the JIRA issue

openshift-examples/keycloak-https.json
```diff
@@ -42,6 +43,41 @@
...
+            "from": "maximum length of 253 lower case alphanumeric characters, -, and .",
+            "value": "",
+            "required": false
+        },
+        {
+            "displayName": "DB Port",
+            "description": "DB port (default ports for H2 = 9123, MYSQL and MARIADB = 3306, POSTGRES = 5432",
+            "name": "DB_PORT",
+            "value": "9123",
+            "required": true
```
stianst: DB_PORT is not required. It will default to the default port by the DB vendor

openshift-examples/keycloak-https.json
```diff
@@ -42,6 +43,41 @@
...
+            "name": "DB_ADDR",
+            "from": "maximum length of 253 lower case alphanumeric characters, -, and .",
+            "value": "",
+            "required": false
+        },
+        {
+            "displayName": "DB Port",
+            "description": "DB port (default ports for H2 = 9123, MYSQL and MARIADB = 3306, POSTGRES = 5432",
+            "name": "DB_PORT",
+            "value": "9123",
```
stianst: Do not set a default value for DB_PORT

openshift-examples/keycloak-https.json
```diff
@@ -42,6 +43,41 @@
...
+            "displayName": "DB ADDR",
+            "description": "Specify the name of the database pod if you are going to use an external DB. The name of the DB pod should not be 'mysql' due to an keycloak internal bug",
+            "name": "DB_ADDR",
+            "from": "maximum length of 253 lower case alphanumeric characters, -, and .",
+            "value": "",
+            "required": false
+        },
+        {
+            "displayName": "DB Port",
+            "description": "DB port (default ports for H2 = 9123, MYSQL and MARIADB = 3306, POSTGRES = 5432",
```
stianst: Description should not list default ports, but rather mention that default DB vendor ports will be used if not specified

openshift-examples/keycloak-https.json
```diff
@@ -42,6 +43,41 @@
...
+            "description": "Name of the DB scheme",
+            "name": "DB_DATABASE",
+            "value": "keycloak",
+            "required": true
+        },
+        {
+            "displayName": "JDBC Parameters",
+            "description": "additional database driver parameters, if you use 'useSSL=true', you manually need to add a certificate",
+            "name": "JDBC_PARAMS",
+            "value": "useSSL=false",
```
stianst: Do not set a value for JDBC_PARAMS. useSSL=false is not supported by all DBs

openshift-examples/keycloak-https.json
```diff
@@ -42,6 +43,41 @@
...
+            "required": true
+        },
+        {
+            "displayName": "Namespace used for DNS discovery",
```
stianst: Please don't move/refactor other elements, it makes it harder to review your PR

openshift-examples/keycloak-https.json
```diff
@@ -233,14 +263,30 @@
                                         "name": "DB_VENDOR",
                                         "value": "${DB_VENDOR}"
                                     },
+                                    {
+                                        "name": "DB_ADDR",
+                                        "value": "${DB_ADDR}.${NAMESPACE}.svc.cluster.local"
```
stianst: You should add .${NAMESPACE}.svc.cluster.local here as that will only work in some situations. You have effectively broken support for connecting to an external DB by doing this.

stianst:
@slaskawi I disagree with what you are saying - there's nothing magical about connecting to a DB and it should in most cases be supported with env values only and not having to extend the image.

openshift-examples/keycloak-https.json
```diff
@@ -8,14 +8,15 @@
             "tags": "keycloak",
             "version": "4.0.0.Beta2",
             "openshift.io/display-name": "Keycloak",
-            "description": "An example Keycloak server with HTTPS"
+            "description": "An example Keycloak server with HTTPS. \n\nNOTE: If you are going to use an external database, you need to manualy inject the environment variables DB_USER and DB_PASSWORD from your database secret."
```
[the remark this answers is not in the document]
cjohn001: true, but where would the right place be?

openshift-examples/keycloak-https.json
```diff
@@ -8,14 +8,15 @@
...
         }
     },
     "parameters": [
         {
             "displayName": "Application Name",
             "description": "The name for the application.",
             "name": "APPLICATION_NAME",
+            "from": "maximum length of 253 lower case alphanumeric characters, -, and .",
```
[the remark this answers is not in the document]
cjohn001: seen this in other templates, hence I added it here for myself, as I sometimes wrongly used upper case letters

openshift-examples/keycloak-https.json
```diff
@@ -42,6 +43,41 @@
             "value": "H2",
             "required": true
         },
+        {
+            "displayName": "DB ADDR",
+            "description": "Specify the name of the database pod if you are going to use an external DB. The name of the DB pod should not be 'mysql' due to an keycloak internal bug",
```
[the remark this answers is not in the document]
cjohn001: well, true, I fill a bug report

openshift-examples/keycloak-https.json
```diff
@@ -42,6 +43,41 @@
...
+            "name": "DB_ADDR",
+            "from": "maximum length of 253 lower case alphanumeric characters, -, and .",
```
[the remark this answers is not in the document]
cjohn001: not sure if it works without, according to description of howto to use with docker

openshift-examples/keycloak-https.json
cjohn001: this caused the issue for me, if not set the init script set it to a wrong value

openshift-examples/keycloak-https.json
cjohn001: than maybe make it an optional empty parameter. I just tested for h2 and mysql

openshift-examples/keycloak-https.json
cjohn001: next time I know better

openshift-examples/keycloak-https.json
cjohn001: sorry, I do not understand what is wrong with it. What would be the correct value?

openshift-examples/keycloak-https.json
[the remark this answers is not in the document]
stianst: We don't have a place to document the images on OpenShift at the moment, so there is not really any right place.

slaskawi:
@stianst @cjohn001 Ok, I took another look into this. Sorry I didn't notice this before but @cjohn001 only exposed parameters, that are already picked by the CLI scripts (see the [code here](https://github.com/jboss-dockerfiles/keycloak/blob/master/server/tools/cli/databases/mysql/change-database.cli#L2)). Having said that, this PR LGTM, except for the comments @stianst already mentioned.

However, we need to think how do we want to support the external DBs. In my opinion a recommended way of doing this is connecting through a `Service` with `Endpoints` pointing to the external DB. You may read about this approach [here](https://docs.openshift.com/container-platform/3.5/dev_guide/integrating_external_services.html). In my opinion it has some advantages over connecting to a DB directly since Kubernetes has some objects to manage the connection and from the Keycloak Pod point of view, we are using the same way as if the DB was deployed in OpenShift/Kube.

If you agree with me, the `DB_ADDR` doesn't have to use `svc.cluster.local` since all Pods in Kubernetes/OpenShift use the internal DNS server to resolve the names. This change will work regardless to the way the DB is set up (internal/external).

cjohn001:
@slaskawi: Ok, I am quite new to the entire Openshift thing. Seems that my addressing will only work if both pods are on the same node. So my solution is suboptimal. As I am not shure how this service discovery thing should work, I think it might be better if we close my pull request and one of you guys, who is more familiar with the topics, proposes a better solution. My unterstanding was, that DB ADDR is the name of the service the database pod provides. However, from your link, I assume, that the service is an extrnal object and that this needs to be referenced rather than the database pod. So please provide a better solution.

cjohn001:
@abstractj : Issue regarding setting of automatic mysql environment variable setup in init script was create https://issues.jboss.org/browse/KEYCLOAK-8532

PR #156, opened 2018-10-05
Fixes KEYCLOAK-8459. Avoid crash when started with more than one network

Nowheresly:
When keycloak is started with several networks, `hostname -i` returns a space separated list of ip addresses. Unfortunately, `jboss.bind.address` and `jboss.bind.address.private` only accept a single ip.
This PR just ensure to loop over each ip to set BIND_OPTS appropriately.


fri-sch:
I just ran into the same problem. I can confirm that this change fixed the problem for me.

PR #157, opened 2018-10-18
Using server config from different params

slaskawi:


server/tools/docker-entrypoint.sh
```diff
@@ -51,9 +51,9 @@ SYS_PROPS+=" $BIND_OPTS"
 # Configuration #
 #################
 
-# If the "-c" parameter is not present, append the HA profile.
-if echo "$@" | egrep -v -- "-c "; then
-    SYS_PROPS+=" -c standalone-ha.xml"
+# If the server configuration parameter is not present, append the HA profile.
+if echo "$@" | egrep -v -- '-c |-c=|--server-config |--server-config='; then
```
sschu: To be safer, one could actually add a " " before every "-" so it is clear that it is a configuration switch. But maybe it is also fine without...

stianst:
Please add a JIRA and provide a description of what you are trying to achieve

0ka:
That PR will solve the issue reported in https://issues.jboss.org/browse/KEYCLOAK-9393

sschu:
What I was trying to achieve is basically to be more robust against the different ways a server configuration can be specified. Originally, the script only checked whether a "-c " parameter was present. However, a server configuration could also be provided using "-c=", "--server-config " or "--server-config=", depending on how the docker entrypoint/CMD looks like. We have the above fix running in our custom image for a while now so it should be safe.

sschu:
Is there a special way to link an issue? Because the issue fixed by this is mentioned in an above comment.

sschu:
@stianst any update on this?

PR #159, opened 2018-11-08
KEYCLOAK-10198 KEYCLOAK-10199: Allow JGroups transport stack configuration

fluescher:
We have a clustered keycloak deployment on Google Kubernetes Engine. Since Google Cloud Platform does not support UDP multicast we use TCP.

The configuration option introduced in this PR greatly simplifies the configuration. The JGroups transport stack can be changed using the environment variable JGROUPS_TRANSPORT_STACK

### DIFF
--- a/server/tools/jgroups.sh
+++ b/server/tools/jgroups.sh
@@ -10,7 +10,7 @@ if [ -n "$JGROUPS_DISCOVERY_PROTOCOL" ]; then
     echo "Setting JGroups discovery to $JGROUPS_DISCOVERY_PROTOCOL with properties $JGROUPS_DISCOVERY_PROPERTIES_PARSED"
     echo "set keycloak_jgroups_discovery_protocol=${JGROUPS_DISCOVERY_PROTOCOL}" >> "$JBOSS_HOME/bin/.jbossclirc"
     echo "set keycloak_jgroups_discovery_protocol_properties=${JGROUPS_DISCOVERY_PROPERTIES_PARSED}" >> "$JBOSS_HOME/bin/.jbossclirc"
-    echo "set keycloak_jgroups_transport_stack=${JGROUPS_TRANSPORT_STACK:-udp}" >> "$JBOSS_HOME/bin/.jbossclirc"
+    echo "set keycloak_jgroups_transport_stack=${JGROUPS_TRANSPORT_STACK:-tcp}" >> "$JBOSS_HOME/bin/.jbossclirc"
     # If there's a specific CLI file for given protocol - execute it. If not, we should be good with the default one.
     if [ -f "/opt/jboss/tools/cli/jgroups/discovery/$JGROUPS_DISCOVERY_PROTOCOL.cli" ]; then
        $JBOSS_HOME/bin/jboss-cli.sh --file="/opt/jboss/tools/cli/jgroups/discovery/$JGROUPS_DISCOVERY_PROTOCOL.cli" >& /dev/null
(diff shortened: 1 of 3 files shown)


slaskawi:
If my memory is correct, there was a parameter in Wildfly to control, which stack to use (we used to use it like this `-Djboss.default.jgroups.stack=tcp`). However, I see this parameter is gone. 

Let me ask my Wildfly friends about this. Maybe there's some parameter to control this behavior supported out of the box. I will get back to as soon as I find something out.

fluescher:
@slaskawi did you get a response from the Wildfly colleagues?

server/tools/jgroups.sh
```diff
@@ -10,6 +10,7 @@ if [ -n "$JGROUPS_DISCOVERY_PROTOCOL" ]; then
     echo "Setting JGroups discovery to $JGROUPS_DISCOVERY_PROTOCOL with properties $JGROUPS_DISCOVERY_PROPERTIES_PARSED"
     echo "set keycloak_jgroups_discovery_protocol=${JGROUPS_DISCOVERY_PROTOCOL}" >> "$JBOSS_HOME/bin/.jbossclirc"
     echo "set keycloak_jgroups_discovery_protocol_properties=${JGROUPS_DISCOVERY_PROPERTIES_PARSED}" >> "$JBOSS_HOME/bin/.jbossclirc"
+    echo "set keycloak_jgroups_transport_stack=${JGROUPS_TRANSPORT_STACK:-udp}" >> "$JBOSS_HOME/bin/.jbossclirc"
```
slaskawi: I'm just thinking whether or not we should TCP as the default stack.... Most of the folks with need it for AWS or GCP deployments.

WDYT?

rhusar: That to me seems orthogonal to this original issue, which is "allowing to change the stack".

rhusar: But to clarify: there seems to be a major performance gain in certain cases we have tested with TCP_NIO2, however, caused certain issues and had to be reverted. However, if you just wait a little longer until we can stabilize TCP_NIO2, you will get the switch out of box. See https://issues.jboss.org/browse/WFLY-8066

slaskawi: My original motivation was to allow folks using this image right away in AWS and GCP. In those deployment scenarios, they would need to switch it to TCP.

But when I think about it, it breaks backwards compatibility. If anyone used our image on prem, he'd need to add a `JGROUPS_TRANSPORT_STACK` environmental variable, whereas the new deployments on AWS would need to use it. So let's keep it as is and maybe someday we'll consider changing the default to TCP.

jkroepke:
Any news about this PR?

would be great to see this upstream inside the docker image.

server/tools/jgroups.sh
slaskawi: After doing a some research and talking with @stianst, I believe the best option is to use TCP by default. Could we incorporate that change into this PR?

fluescher: Sure, we just need to make sure the new behaviour is mentioned in the release notes and documentation.

gpguerra:
I was having problem to deploy a keycloak cluster on Oracle cloud container engine (OKE).
The nodes can find themselves but can't comunicate (timeouts). So I had to edit standalone-ha.xml to use tcp as default 
`<channel name="ee" stack="tcp" cluster="ejb"/>` 
and now it works. But its a bit ugly.

fluescher:
I will change the default to TCP in this PR tomorrow. Is there anything else that needs to be done in order to get it merged?

fluescher:
Could you have a look @slaskawi?

fluescher:
Hi @slaskawi Thanks for having a look! I added the JIRA references accordingly.

envy7:
Hey @stianst I see this code was pushed 14 days ago, but when pulling the latest docker image of keycloak from dockerhub, it is not pulling these changes, can you please check regarding the same
I am using the jboss/keycloak:latest image, should I pulling some other image?

stianst:
Images are updated when we do releases, not continuously.

redbaron:
Word of warning, we see significant latency increase on cluster membership change when running in TCP mode. It is being discussed here: https://issues.redhat.com/browse/KEYCLOAK-13310

PR #160, opened 2018-11-08
Fixing truststore import message

brix4dayz:
I was using the `X509_CA_BUNDLE` feature when I noticed in the logs that it `"Failed to import certificates from system's Java CA certificate bundle into Keycloak truststore!"`.

I pulled the container down and built the truststore manually like the script does, and it succeeded. I checked the return code and it was 0, so I believe this is backwards.


slaskawi:
Very good eye @brix4dayz! Thanks for fixing this!

@iankko I think we have the same type of mistake in the downstream...

PR #162, opened 2018-11-13
server: Fix installation when hostname -i returns multiple IPs

stlaz:
`hostname -i` can return multiple IPs and its use is discouraged by
hostname's man. Use `hostname -I` and get its first IP.

### DIFF
--- a/server/tools/docker-entrypoint.sh
+++ b/server/tools/docker-entrypoint.sh
@@ -40,10 +40,7 @@ if [ -z "$BIND" ]; then
     BIND=$(hostname -i)
 fi
 if [ -z "$BIND_OPTS" ]; then
-    for BIND_IP in $BIND
-    do
-        BIND_OPTS+=" -Djboss.bind.address=$BIND_IP -Djboss.bind.address.private=$BIND_IP "
-    done
+    BIND_OPTS="-Djboss.bind.address=$BIND -Djboss.bind.address.private=$BIND"
 fi
 SYS_PROPS+=" $BIND_OPTS"
 
--- a/server/tools/docker-entrypoint.sh
+++ b/server/tools/docker-entrypoint.sh
@@ -36,9 +36,27 @@ fi
 # JGroups bind options #
 ########################
 
+# takes an IP address to be checked
+# returns "false" if not a valid IPv4, nothing otherwise
+function is_valid_ipv4() {
+    local address=$1
+    local ipv4_regex="^((25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.){3}(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)$"
+    [[ "$address" =~ $ipv4_regex ]] || echo -n "false"
+}
+
 if [ -z "$BIND" ]; then
-    BIND=$(hostname -i)
+    # Get an IPv4 address from the interfaces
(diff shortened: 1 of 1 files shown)


stlaz:
I found this bug while trying to run keycloak container with podman on one of my machines `podman run -p 8080:8080 jboss/keycloak:4.5.0.Final`.

The installation would eventually fail with `WFLYSRV0073: Invalid option '10.88.0.2'` since `hostname -i` stuck `fe80::1837:82ff:feed:ef73%eth0 10.88.0.11` in the `BIND` variable.

slaskawi:
@stlaz Ok, I found a commit trying to fix a similar thing: https://github.com/jboss-dockerfiles/keycloak/commit/89c77d05e70d95826806ea658f73fb5270c32f4d (sent by @Nowheresly). 

Now, I realized that adding multiple addresses to the `BIND_OPTS` will not work. It seems that Wildfly will interpret only the first one. Therefore, I think we should just leave your part:

```
BIND=$(hostname -I | cut -d' ' -f1)
```

and remove the part implemented by @Nowheresly. 

I guess we have at least two options how to do it:
- revert `89c77d05e70d95826806ea658f73fb5270c32f4d` and proceed with your changes
- remove the loop part in your commit. 

Which one to use, I'll leave to you. I have a small preference towards to the first one. 

I also recommend to enhance the script to look for the IPv4 pattern (not just blindly grab the first or the second part of an `hostname -i` output).

stlaz:
@slaskawi good find!

The first IPv4 will now be stored in the BIND variable. It could, however, leave it blank if no IPv4 is found among the interface addresses - I wonder whether we'd like to check such cases.

Nowheresly:
> Now, I realized that adding multiple addresses to the `BIND_OPTS` will not work. It seems that Wildfly will interpret only the first one. 

bad news... this information seems indeed hard to find in the Wildfly documentation.

I'm just wondering if picking just one of the IPs is enough. Basically, our keycloak instance has 2 IPs because our db is in a dedicated network while the applications that need to contact keycloak are in another network.

My guess (but maybe I'm wrong) is that, in such case, we should bind the IP corresponding of the applications networks, not the one of the db network.

Do you think wildfly will behave the same if we bind any of the IP indifferently?

slaskawi:
@Nowheresly It would be really helpful if you could confirm or deny that Wildfly has bounded to two IPs in your case. Personally, I don't think so but it would be good to check on your case. If I'm right, and Wildfly picked only one of them, could you please give us a hint which one it picked (the former or the latter)?

slaskawi:
@stlaz If we leave it blank, Wildfly will bind to `127.0.0.1` (this is the default in the configuration). If no valid IP is found, perhaps we should print out a warning?

stlaz:
@slaskawi added a warning message. Did not mention the loopback being used in case Wildfly changes its behavior, but can append the message about it if you want

slaskawi:
@stlaz I believe I fixed this one as a part of https://github.com/jboss-dockerfiles/keycloak/pull/186. 

Could you please check if my PR fixes your problem as well (and if it doesn, please close this PR).

slaskawi:
I believe we can close this one @stianst

stlaz:
Yeah, sorry, I haven't had much time to test this. I believe your judgment on the issue, closing.

PR #168, opened 2018-11-21
Allow Switching the discovery Stack to TCP

rhnewtron:
On Openshift we need "multicast" for the default of dns.DNS_PING to
work with UDP . This may not be available in older clusters (I think a
3.6 feature) and also it may be a security problem in larger, shared
projects.

### DIFF
--- a/server/tools/cli/jgroups/discovery/default.cli
+++ b/server/tools/cli/jgroups/discovery/default.cli
@@ -1,5 +1,8 @@
 embed-server --server-config=standalone-ha.xml --std-out=echo
 batch
+
+/subsystem=jgroups/channel=ee:write-attribute(name=stack,value=${env.JGROUPS_DISCOVERY_STACK:udp})
+
 /subsystem=jgroups/stack=udp/protocol=PING:remove()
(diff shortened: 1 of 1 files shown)


slaskawi:
This PR overlaps with https://github.com/jboss-dockerfiles/keycloak/pull/159, which slightly more flexible. I'm planning to review it today so I guess we should close this one.

However, thanks a lot for the contribution. We appreciate you folks looking into this!

rhnewtron:
Great, does it make sense to open a pull request which adds support to add custom ".cli" scripts via configmap ? I think in general cusomization is missing right now.

Also where can I find the source of the dockerhub readme file for the image ? The documentation os really, how should I say - "limited" in regards to clustering. It should mention that we need a "headless" service for DNS_PING and it should also mention, that you have to enable multicast for UDP to work.

slaskawi:
@rhnewtron I believe the functionality you asked for has been merged here: https://github.com/jboss-dockerfiles/keycloak/pull/176

If the solution from that PR satisfies you, could you please close this PR?

slaskawi:
I believe we can close this one @stianst

PR #169, opened 2018-12-06
[KEYCLOAK-9017] Add option to set database schema to container images

bartmentech:
This will fix https://issues.jboss.org/browse/KEYCLOAK-9017

### DIFF
--- a/docker-compose-examples/keycloak-postgres.yml
+++ b/docker-compose-examples/keycloak-postgres.yml
@@ -6,13 +6,14 @@ volumes:
 
 services:
   postgres:
-      image: postgres
+      build: ./postgres
       volumes:
         - postgres_data:/var/lib/postgresql/data
       environment:
         POSTGRES_DB: keycloak
         POSTGRES_USER: keycloak
         POSTGRES_PASSWORD: password
+        POSTGRES_SCHEMA: keycloak
   keycloak:
       image: jboss/keycloak
       environment:
@@ -20,6 +21,7 @@ services:
         DB_ADDR: postgres
         DB_DATABASE: keycloak
         DB_USER: keycloak
+        DB_SCHEMA: keycloak
         DB_PASSWORD: password
         KEYCLOAK_USER: admin
         KEYCLOAK_PASSWORD: Pa55w0rd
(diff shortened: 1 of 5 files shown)


stianst:
Looks like this is only tested with PostgreSQL, we need to consider how it affects other DBs. The property you're setting on the connectionsJpa provider does affect all DBs afaik

docker-compose-examples/keycloak-postgres.yml
```diff
@@ -6,20 +6,22 @@ volumes:
 
 services:
   postgres:
-      image: postgres
+      build: ./postgres
```
stianst: Remove this change. We should not require users to build a Postgres image to use the example. I'd say the schema should be set to public here and it's up to the users to figure out how to add a schema to Postgres and not something we should cover on our end.

docker-compose-examples/postgres/Dockerfile
```diff
@@ -0,0 +1,4 @@
+FROM postgres:11
```
stianst: Remove this file - see above comment

docker-compose-examples/postgres/create_schema.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
```
stianst: Remove this file - see above comment

server/README.md
```diff
@@ -72,6 +72,7 @@ Generic variable names can be used to configure any Database type, defaults may
 - `DB_ADDR`: Specify hostname of the database (optional)
 - `DB_PORT`: Specify port of the database (optional, default is DB vendor default port)
 - `DB_DATABASE`: Specify name of the database to use (optional, default is `keycloak`).
+- `DB_SCHEMA`: Specify name of the schema to use, this is only used when PostgreSQL is used (optional, default is `public`).
```
stianst: Change to:

Specify name of the schema to use for DB that support schemas (optional, default is `public` on Postgres).

bartmentech:
Thank you for your comments. I will update it this week

stianst:
Thanks

PR #171, opened 2018-12-28
Updated Dockerfile for Gatekeeper

lukasmrtvy:
Can anyone care of this Dockerfile ? 

I guess that images are pushed here `https://hub.docker.com/r/keycloak/keycloak-gatekeeper`, right?

### DIFF
--- a/gatekeeper/Dockerfile
+++ b/gatekeeper/Dockerfile
@@ -1,7 +1,7 @@
-FROM alpine:3.7
+FROM alpine:3.8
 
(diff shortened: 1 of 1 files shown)


stianst:
Thanks for your PR.

KC versions are updated when we do releases, see https://github.com/jboss-dockerfiles/keycloak/blob/4.8.1.Final/gatekeeper/Dockerfile.

Due to restrictions on github we can't yet automate updating version in master, hence why it's sometimes forgotten.

PR #173, opened 2019-01-22
KEYCLOAK-9881: add postgres connection fail-over url

malmuzaini:
jdbc allow connection fail-over for postgres. Please refer to https://jdbc.postgresql.org/documentation/head/connect.html

### DIFF
--- a/server/README.md
+++ b/server/README.md
@@ -69,7 +69,10 @@ If the DB can't be detected it will default to the embedded H2 database.
 
 Generic variable names can be used to configure any Database type, defaults may vary depending on the Database.
 
-- `DB_ADDR`: Specify hostname of the database (optional)
+- `DB_ADDR`: Specify hostname of the database (optional). For postgres only, you can provide a list of hostnames separated by
+comma to failover alternative host. The hostname can be the host only or pair of host and port, as example host1,host2 or 
+host1:5421,host2:5436 or host1,host2:5000. And keycloak will append DB_PORT (if specify) to the hosts without port, 
+otherwise it will append the default port 5432, again to the address without port only.
 - `DB_PORT`: Specify port of the database (optional, default is DB vendor default port)
 - `DB_DATABASE`: Specify name of the database to use (optional, default is `keycloak`).
 - `DB_USER`: Specify user to use to authenticate to the database (optional, default is `keycloak`).
--- a/server/tools/cli/databases/postgres/change-database.cli
+++ b/server/tools/cli/databases/postgres/change-database.cli
@@ -1,5 +1,5 @@
 /subsystem=datasources/data-source=KeycloakDS: remove()
-/subsystem=datasources/data-source=KeycloakDS: add(jndi-name=java:jboss/datasources/KeycloakDS,enabled=true,use-java-context=true,use-ccm=true, connection-url=jdbc:postgresql://${env.DB_ADDR:postgres}:${env.DB_PORT:5432}/${env.DB_DATABASE:keycloak}${env.JDBC_PARAMS:}, driver-name=postgresql)
+/subsystem=datasources/data-source=KeycloakDS: add(jndi-name=java:jboss/datasources/KeycloakDS,enabled=true,use-java-context=true,use-ccm=true, connection-url=jdbc:postgresql://${env.DB_ADDR:postgres}/${env.DB_DATABASE:keycloak}${env.JDBC_PARAMS:}, driver-name=postgresql)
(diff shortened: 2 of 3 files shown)


stianst:
Adding yet another DB vendor (postgres-ha) that is configured differently is not the correct approach here. Would be better to allow the DB_HOST to be a list. Makes it a bit tricky with DB_PORT if different hosts have different ports. Having that as a separate variable was probably a mistake from the start.

I'd suggest allowing DB_HOST to contain a comma separated list and also allowing including the port. So valid values could be:

DB_HOST=host1
DB_HOST=host1,host2
DB_HOST=host1:5643,host2:6757

We should make DB_PORT optional and if it is added then it's added to all hosts.

Please also create a corresponding JIRA for this PR.

stianst:

malmuzaini:
I opened this issue KEYCLOAK-9881.
I totally agree with your suggestion let me fix it.

malmuzaini:
@stianst Can you please reopen the PR!

malmuzaini:
> Adding yet another DB vendor (postgres-ha) that is configured differently is not the correct approach here. Would be better to allow the DB_HOST to be a list. Makes it a bit tricky with DB_PORT if different hosts have different ports. Having that as a separate variable was probably a mistake from the start.
> 
> I'd suggest allowing DB_HOST to contain a comma separated list and also allowing including the port. So valid values could be:
> 
> DB_HOST=host1
> DB_HOST=host1,host2
> DB_HOST=host1:5643,host2:6757
> 
> We should make DB_PORT optional and if it is added then it's added to all hosts.
> 
> Please also create a corresponding JIRA for this PR.

The hosts along with their ports should be passed as a single value, like this:
DB_ADDR=host1:5643,host2:6757
So the format should be host:port and if there is more than one host they should be separated by comma.
DB_PORT will be removed for postgres. So even if it's passed it will be ignored in case the driver is postgres.

What your say?

stianst:
We can't just remove DB_PORT as folks may depend on it. We also need to keep this consistent across all DBs.

DB_PORT should be marked as deprecated with the new approach setting the port as part of DB_HOST. If DB_PORT is set it can be appended to DB_HOST in the case DB_HOST contains a single host and does not have the port set. That is probably the simplest option that is still backwards compatible.

malmuzaini:
I am aware of this fact.

You said " If DB_PORT is set it can be appended to DB_HOST in the case DB_HOST contains a single host and does not have the port set. That is probably the simplest option that is still backwards compatible."

This is a logic, and it's not a straight one, so it's hard to implement in the cli file. So I am thinking to implement it in docker-entrypoint.sh, what your say @s01ipsist

malmuzaini:
@stianst Could you please reply

stianst:
Sorry for lack of response, missed this one. Yes, logic can't be handled in CLI so would need to be handled in docker-entrypoint.sh

slaskawi:
@stianst @malmuzaini What is the status of this PR? As far as I can tell from the comments, it is ready to be merged.

stianst:
This is not ready to be merged as it needs to incorporate the changes that was discussed first.

malmuzaini:
Allow me to the end of this weak to apply the changes as discussed.

malmuzaini:
@stianst @slaskawi Please review

stianst:
Approach looks good to me. Haven't had a chance to try it out yet though. Perhaps @slaskawi could do that?

Also, wondering if this shouldn't also be applied to all DBs? As far as I know it's common to have a comma separate list in the JDBC URI. At least MySQL does that:
https://dev.mysql.com/doc/connector-j/5.1/en/connector-j-config-failover.html

I know it's out of the scope of what you wanted, but would be a nice addition.

malmuzaini:
@stianst Thank you. I agree since our concern was postgres, I opened the issue for postgres. But as you said it's possible for the other DBs. I am willing to do it for the other DBs, but is it ok to but it in this PR?

malmuzaini:
@stianst &  @slaskawi could you please look into this PR

slaskawi:
@malmuzaini @stianst This looks fine to me. I'm OK with merging it as is.

As @stianst, it would be great to provide this kind of functionality for other DBs as well. Whether it should be done in this PR or in a separate - it's up to you, @malmuzaini.

Thanks for a great work!

malmuzaini:
@slaskawi @stianst I prefer to create a new one one for the others. So let's merge this. 🙌🏽

malmuzaini:
@slaskawi @stianst Plz merge this

PR #175, opened 2019-02-07
[KEYCLOAK-8310] added environment variable to be able to use the 'alw…

wesley9620:
Added support for [KEYCLOAK-8310].  I added an environment variable KEYCLOAK_ALWAYS_HTTPS, modified the docker-entrypoint.sh and the hostname.cli


server/tools/docker-entrypoint.sh
```diff
@@ -22,6 +22,10 @@ if [ "$KEYCLOAK_HOSTNAME" != "" ]; then
     if [ "$KEYCLOAK_HTTPS_PORT" != "" ]; then
         SYS_PROPS+=" -Dkeycloak.hostname.fixed.httpsPort=$KEYCLOAK_HTTPS_PORT"
     fi
+
+    if [ "$KEYCLOAK_ALWAYS_HTTPS" != "" ]; then
+            SYS_PROPS+=" -Dkeycloak.hostname.fixed.alwaysHttps=$KEYCLOAK_ALWAYS_HTTPS"
```
slaskawi: @stianst Shouldn't we always set that when using HTTPS?

abstractj: @slaskawi +1

stianst: @slaskawi Not sure what you mean.

slaskawi: Shouldn't we always add `-Dkeycloak.hostname.fixed.alwaysHttps=$KEYCLOAK_ALWAYS_HTTPS` if `KEYCLOAK_HTTPS_PORT` is not empty?

stianst: No, because some folks may want both http and https. For instance they may have a reverse proxy that uses http, but internal clients use https.

stianst:
@slaskawi the changes look good to me, but haven't had a chance to test it. Can you give it a go?

exaV:
Have you had a chance to test this feature yet? We need it and currently use a workaround that is rather ugly.

PR #176, opened 2019-02-07
Support a docker-entrypoint.d directory

wouterhund:
This change will allow extending images to simply drop files in `/opt/jboss/keycloak/tools/docker-entrypoint.d` to have them ran on container startup. Special support for `.cli` files is provided.

This is useful for, in our case, simple tweaks to the `standalone.xml`.

### DIFF
--- a/server/README.md
+++ b/server/README.md
@@ -166,12 +166,16 @@ directory.
 
 ## Running custom scripts on startup
 
-To run a custom `.cli` script on container startup extend the Keycloak image and add the `.cli` file to the
-`/opt/jboss/keycloak/tools/docker-entrypoint.d` directory (create it if necessary).
+To run custom scripts on container startup either extend the Keycloak image or use a volume, then add your script to
+the `/docker-entrypoint.d` directory.
 
-It is also possible to add executable scripts (`chmod +x`) to this directory.
+Two types of scripts are supported:
 
-Files are ran in alphabetical order.
+* WildFly `.cli` [scripts](https://docs.jboss.org/author/display/WFLY/Command+Line+Interface)
+
+* Any executable (`chmod +x`) script
+
+Scripts are ran in alphabetical order.
 
 ## Clustering
 
--- a/server/README.md
+++ b/server/README.md
@@ -164,6 +164,14 @@ To set the welcome theme, use the following environment value :
 To add a custom provider extend the Keycloak image and add the provider to the `/opt/jboss/keycloak/standalone/deployments/`
 directory.
 
+## Running custom scripts on startup
+
+To run a custom `.cli` script on container startup extend the Keycloak image and add the `.cli` file to the
+`/opt/jboss/keycloak/tools/docker-entrypoint.d` directory (create it if necessary).
+
+It is also possible to add executable scripts (`chmod +x`) to this directory.
+
+Files are ran in alphabetical order.
 
 ## Clustering
 
--- a/server/README.md
+++ b/server/README.md
@@ -166,8 +166,9 @@ directory.
 
 ## Running custom scripts on startup
 
-To run custom scripts on container startup either extend the Keycloak image or use a volume, then add your script to
-the `/docker-entrypoint.d` directory.
+**Warning**: Custom scripts have no guarantees. The directory layout within the image may change at any time.
+
+To run custom scripts on container startup place a file in the `/opt/jboss/startup-scripts` directory.
 
 Two types of scripts are supported:
 
@@ -176,6 +177,23 @@ Two types of scripts are supported:
 * Any executable (`chmod +x`) script
 
 Scripts are ran in alphabetical order.
+
+### Adding custom script using Dockerfile
+
+A custom script can be added by creating your own `Dockerfile`:
+
+```
+FROM keycloak
+COPY custom-scripts/ /opt/jboss/startup-scripts/
+```
+
+### Adding custom script using volumes
+
+A single custom script can be added as a volume: `docker run -v /some/dir/my-script.cli:/opt/jboss/startup-scripts/my-script.cli`
+Or you can volume the entire directory to supply a directory of scripts.
+
+Note that when combining the approach of extending the image and `volume`ing the entire directory, the volume will override
+all scripts shipped in the image.
 
 ## Clustering
 
--- a/server/tools/autorun.sh
+++ b/server/tools/autorun.sh
@@ -1,7 +1,7 @@
 #!/bin/bash -e
 cd /opt/jboss/keycloak
 
-ENTRYPOINT_DIR=/docker-entrypoint.d
+ENTRYPOINT_DIR=/opt/jboss/startup-scripts
 
 if [[ -d "$ENTRYPOINT_DIR" ]]; then
   # First run cli autoruns
--- a/server/tools/autorun.sh
+++ b/server/tools/autorun.sh
@@ -1,0 +1,19 @@
+#!/bin/bash
+cd /opt/jboss/keycloak
+
+ENTRYPOINT_DIR=/opt/jboss/tools/docker-entrypoint.d
+
+if [[ -d "$ENTRYPOINT_DIR" ]]; then
+  # First run cli autoruns
+  for f in "$ENTRYPOINT_DIR"/*; do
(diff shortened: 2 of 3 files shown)


slaskawi:
I'm not convinced we would like to have this kind of functionality in our image.

After all, if you'd like to execute a custom script, you need to extend our image. And since you need to extend it, you can provide your own bootstrap file.

However, I do agree that the current image structure doesn't allow to execute a custom script just before running the `standalone.sh` script. One of the ways would be to let `docker-entrypoint.sh` file configure the instance and then run it manually. The `ENTRYPOINT` of a docker file could look like `ENTRYPOINT [ "/opt/jboss/tools/docker-entrypoint.sh", "/opt/jboss/keycloak/bin/standalone.sh $SYS_PROPS" ]` (of course with some adjustments, I just wanted to present the idea). 

With such an approach, everyone could use his own scripts between executing `docker-entrypoint.sh` and `standalone.sh`. 

@stianst @abstractj WDYT?

wouterhund:
It is fairly common for images to provide a `docker-entrypoint.d` directory. It would be nice if the `keycloak` did as well. Currently it's not really possible to extend the startup procedure of this image without some relatively complicated changes or using `patch`. It should be as simple as simply dropping a file in a folder.

I don't really understand your `ENTRYPOINT` suggestion, docker does not support multiple entrypoints as far as I'm aware. You'd have to do some shell scripting for that, which somewhat defeats the point of simplicity.

Among others, the official database images also provide similar functionality:

* https://hub.docker.com/_/postgres (`docker-entrypoint-initdb.d`)

* https://hub.docker.com/_/mysql (`docker-entrypoint-initdb.d`)

wouterhund:
Also, consider multiple images extending each other all adding a piece of customization. I don't think that would be possible with your suggestion. It is easy and natural with a `docker-entrypoint.d` directory.

abstractj:
@wouterhund even though I understand your point and see that projects like `postgres` or `mysql` are using it. That's not the standard across the Docker images for JBoss (https://github.com/jboss-dockerfiles).

Here's my suggestion:
* Send an e-mail to keycloak-dev with your proposal. Because it seems like a super simple change, but it affects our images. Also, you gonna reach a broader audience.
*  File a Jira, so we can prioritize and include in further releases.

slaskawi:
I totally agree with @abstractj. This probably deserves some discussion.

wouterhund:
We'll keep maintaining our custom patched docker image for now then, since we need a way to easily execute scripts on startup. Shame this won't be merged upstream, the patch has been tested, is functional and backwards compatible.

Are you planning to solve this problem differently within the "JBoss standard"?

abstractj:
Hi @wouterhund if there's anything has to be solved an enhancement or a feature request. I strongly encourage you to discuss with us at keycloak-dev mailing list.

slaskawi:
I triggered the discussion here: http://lists.jboss.org/pipermail/keycloak-dev/2019-February/011642.html

llech:
As for now, the only possibility to plug-in before keycloak start is seding the docker-entrypoint script, which is ofter described as write-only programming. And it is not resistant towards big changes in that file. 

Supporting custom scripts to run before calling standalone.sh would be a great feature.

server/README.md
```diff
@@ -164,6 +164,14 @@ To set the welcome theme, use the following environment value :
 To add a custom provider extend the Keycloak image and add the provider to the `/opt/jboss/keycloak/standalone/deployments/`
 directory.
 
+## Running custom scripts on startup
```
stianst: Add to documentation that scripts can be added by extending image or by mounting.

server/README.md
```diff
@@ -164,6 +164,14 @@ To set the welcome theme, use the following environment value :
...
+
+To run a custom `.cli` script on container startup extend the Keycloak image and add the `.cli` file to the
+`/opt/jboss/keycloak/tools/docker-entrypoint.d` directory (create it if necessary).
```
stianst: Change name of the directory. Something like "/opt/jboss/startup-scripts". "tools" shouldn't be there and it definitively should not contain the name "docker".

wouterhund: Why definitely not the name docker? The name `docker-entrypoint.d` is the community standard as I've shown in my previous comments.

wouterhund:
FWIW, I've enabled contributors to directly edit this PR, so feel free to make the changes you wish to see. This might be more convenient for doc changes.

server/README.md
slaskawi: There are quite a few arguments why we don't like to include "docker" in the description...

One of them is that we want to make Keycloak as friendly as possible for the Kubernetes-based environments (such as OpenShift). Since Kubernetes can work without Docker (using Podman), it creates a sort of conflict in our minds. Not to mention, that you can build Keycloak image without Docker as well (using Buildah). 

So I agree with @stianst here, please pick some more neutral name. Using `/opt/jboss/startup-scripts` seems to be a good start.

server/README.md
slaskawi: I believe this hasn't been properly resolved yet. 

I would like to request some additional examples that demonstrate how to mount such scripts into the image (with `docker -v` option).

server/README.md
```diff
@@ -164,6 +164,18 @@ To set the welcome theme, use the following environment value :
 To add a custom provider extend the Keycloak image and add the provider to the `/opt/jboss/keycloak/standalone/deployments/`
 directory.
 
+## Running custom scripts on startup
+
+To run custom scripts on container startup either extend the Keycloak image or use a volume, then add your script to
```
slaskawi: As I mentioned on the mailing list (see http://lists.jboss.org/pipermail/keycloak-dev/2019-February/011642.html) we can't guarantee any stability of this scripts. We may change the location of the Keycloak distribution to some other directory for example and it is very likely that user script will break. We may also change the way the distribution itself is structured and that will also cause such scripts to fail.

I would explicitly like to make this statement part of the contract. In other words, please mention in the documentation that users who use this functionality do it without any guarantees.

slaskawi:
> FWIW, I've enabled contributors to directly edit this PR, so feel free to make the changes you wish to see. This might be more convenient for doc changes.

Allowing the contributors to edit works really great for minor tweaking (like a typo for example). It doesn't work quite well with some major changes (as we requested on this PR). 

Even though, the review and feedback process takes quite a lot of time, I highly encourage you to introduce this changes and get this functionality merged. The idea is really great and many folks are waiting for it (BTW, we received very good feedback from the mailing list!).

server/README.md
wouterhund: Perhaps the `docker-entrypoint.sh` (already existing, part of the reason I chose this name) should also be renamed?

server/README.md
wouterhund: This better?

server/README.md
wouterhund: I've relocated the directory to `/opt/jboss/startup-scripts`.

server/README.md
slaskawi: Yes, thank you!

server/README.md
slaskawi: Very good point! I've created [KEYCLOAK-9609](https://issues.jboss.org/browse/KEYCLOAK-9609) to replace `docker-entrypoint.sh` with better name. Thanks for pointing this out!

server/tools/autorun.sh
```diff
@@ -0,0 +1,19 @@
...
+  # First run cli autoruns
+  for f in "$ENTRYPOINT_DIR"/*; do
+    if [[ "$f" == *.cli ]]; then
+      echo "Executing cli script: $f"
+      bin/jboss-cli.sh --file="$f"
+    elif [[ -x "$f" ]]; then
+      echo "Executing: $f"
+      "$f"
+    else
+      echo "Ignoring file in docker-entrypoint.d directory (not *.cli or executable): $f"
```
oscar-b: This should be changed to mention "startup-scripts" instead, I guess.

oscar-b:
Any idea when this will be pushed to a tagged image? Took me a while to figure out why it didn't work :|

server/tools/autorun.sh
wouterhund: I've made #178 for this

PR #177, opened 2019-02-12
Remove dead links from README.md

marceloboeira:
It looks like mysql/pgsql images were merged into the main one and it's likely that's the reason why those links are 404s now


marceloboeira:
@slaskawi can you merge it?

slaskawi:
@marceloboeira Unfortunately I don't have merge permissions. 

@stianst Could you please do the honors?

PR #179, opened 2019-02-24
Update JDBC_POSTGRES_VERSION for sslmode=require

zaclittleberry:
Update the patch version of JDBC_POSTGRES_VERSION from 42.2.2 to 42.2.5 so that keycloak can support sslmode=require as a value in JDBC_PARAMS.


stianst:
@slaskawi can you review and test this please?

zaclittleberry:
@stianst , @slaskawi , should I change the merge target branch to something other than master? I see this request now has merge conflicts with master since master has moved onto the development of Keycloak v5. 

It's probably a good idea to also update JDBC_POSTGRES_VERSION for v5, but I presume v4 will be supported for some time more and that is my primary concern as that is what is stable and running in my case. 

Willing to make the pull request modifications in any which way they are required. Just let me know. Thanks!

stianst:
Not sure why it's showing a merge conflict as you haven't changed the line that has changed. If you just do a rebase on your branch it should be fine afterwards.

PR #182, opened 2019-03-02
KEYCLOAK-9715 add KEYCLOAK_DEFAULT_THEME to set global default theme

mcleane:
Providing the ability to set the global default theme within a container image by setting an environment variable KEYCLOAK_DEFAULT_THEME with the name of the custom theme.

If environment variable is not set, theme will default to keycloak. Same behaviour as the KEYCLOAK_WELCOME_THEME.

Changes tested by :
1. extending container image with custom theme and setting environment variable. Launch container and verify custom theme is set. 
2. executing base container to ensure default is still keycloak theme.

### DIFF
--- a/server/README.md
+++ b/server/README.md
@@ -158,6 +158,9 @@ To set the welcome theme, use the following environment value :
 
 * `KEYCLOAK_WELCOME_THEME`: Specify the theme to use for welcome page (must be non empty and must match an existing theme name)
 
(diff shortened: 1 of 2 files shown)


stianst:
Thanks

ivanstefko:
Hi guys, 
   thanks for that fix. Could you tell me when do you plan to merge these changes to latest branch (when these changes will be included in official release). 

Thanks!

PR #186, opened 2019-04-02
KEYCLOAK-9767 Bind to proper interface

slaskawi:
https://issues.jboss.org/browse/KEYCLOAK-9767

This commit allows running the Keycloak image with Podman. The root cause, why it didn't work was that we were binding to a wrong interface - `fe80::5003:8eff:fefa:3e53%tap0`. I corrected this with this PR.

You may test this using `podman run slaskawi/keycloak-tmp`.

@Nowheresly - This PR may affect you as you worked previously on https://github.com/jboss-dockerfiles/keycloak/pull/156. Could you please verify if it works for you?

@stianst - Could you please test it with your environment? for a short test, please use `podman run slaskawi/keycloak-tmp`. For a longer one, you'd need to rebuild the image.

### DIFF
--- a/server/tools/docker-entrypoint.sh
+++ b/server/tools/docker-entrypoint.sh
@@ -62,7 +62,7 @@ fi
 ########################
 
 if [ -z "$BIND" ]; then
-    BIND=$(hostname -i)
+    BIND=$(hostname --all-ip-addresses)
 fi
 if [ -z "$BIND_OPTS" ]; then
     for BIND_IP in $BIND


stianst:
Works for me, but as I mentioned to you on Slack I'm a bit worried that it won't work for everyone/everywhere. Perhaps drop a mail to keycloak-dev and keycloak-user to ask for some feedback from the community on this one?

slaskawi:
@stianst Email sent. Let's wait a bit for feedback.

server/tools/docker-entrypoint.sh
```diff
@@ -62,7 +62,7 @@ fi
 ########################
 
 if [ -z "$BIND" ]; then
-    BIND=$(hostname -i)
+    BIND=$(hostname -i | tr " " "\n" | egrep '([[:digit:]]{1,3}\.){3}[[:digit:]]{1,3}')
```
hmlnarik: Is it certain that IPv4 addresses would be always available?
Note that only one of the addresses is going to be used since the system property can be set only to one value (iteration on line 70 would pass multiple ones but only a single one would be transferred to system properties)

slaskawi: This is a very good question...

So technically, it is possible to use IPv6 with Docker (see [manual](https://docs.docker.com/v17.09/engine/userguide/networking/default_network/ipv6/#using-ndp-proxying)). However, that would require some configuration changes, as `standalone-ha.xml` declares default addresses like that:

```xml
<socket-binding name="jgroups-mping" interface="private" port="0" multicast-address="${jboss.default.multicast.address:230.0.0.4}" multicast-port="45700"/>
<socket-binding name="jgroups-tcp" interface="private" port="7600"/>
<socket-binding name="jgroups-udp" interface="private" port="55200" multicast-address="${jboss.default.multicast.address:230.0.0.4}" multicast-port="45688"/>
```

In addition to that, the `keycloak/bin/standalone.conf` adds this line:

```
keycloak/bin/standalone.conf:   JAVA_OPTS="-Xms64m -Xmx512m -XX:MetaspaceSize=96M -XX:MaxMetaspaceSize=256m -Djava.net.preferIPv4Stack=true"
```

So technically, it should be possible to run Keyclaok (or Wildfly) with IPv6, but it would require major tweaking (supplying `jboss.default.multicast.address` property and `JAVA_OPTS` environmental variable at least). 

Let me ask @belaban @rhusar and @pferraro if they know any IPv6 use cases for Infinispan or Wildfly. According to my knowledge, we use only IPv4 in containers, Kubernetes or OpenShift environment.

belaban: I think it should be possible to use IPv6 in containers, but most images I know of provide only one (or two) IPv4 addresses.

Nowheresly:
Hi @slaskawi .
Thanks for pointing me out to this PR.
I tested it and unfortunately, it seems it does work for me.

With my settings, keycloak is started in a docker swarm cluster and linked to 3 different networks (overlay driver).

Please have a look at the output of these commands:
```
hostname -i | tr " " "\n" | egrep '([[:digit:]]{1,3}\.){3}[[:digit:]]{1,3}'
10.0.24.29
10.0.14.25
10.0.10.5
```

```
ps -elf
0 S jboss       264      8 19  80   0 - 606063 futex_ 11:17 ?       00:00:50 /usr/lib/jvm/java/bin/java -D[Standalone] -server -Xms64m -Xmx512m -XX:+UnlockExperimentalVMOptions -XX:+UseCGroupMemoryLimitForHeap -XX:MetaspaceSize=96M -XX:MaxMetaspaceSize=256m -Djava.net.preferIPv4Stack=true -Djboss.modules.system.pkgs=org.jboss.byteman -Djava.awt.headless=true -Dorg.jboss.boot.log.file=/opt/jboss/keycloak/standalone/log/server.log -Dlogging.configuration=file:/opt/jboss/keycloak/standalone/configuration/logging.properties -jar /opt/jboss/keycloak/jboss-modules.jar -mp /opt/jboss/keycloak/modules org.jboss.as.standalone -Djboss.home.dir=/opt/jboss/keycloak -Djboss.server.base.dir=/opt/jboss/keycloak/standalone -Djboss.bind.address=10.0.24.29 -Djboss.bind.address.private=10.0.24.29 -Djboss.bind.address=10.0.14.25 -Djboss.bind.address.private=10.0.14.25 -Djboss.bind.address=10.0.10.5 -Djboss.bind.address.private=10.0.10.5 -c=standalone-ha.xml

```

_Expected behavior_:

keycloak can be reachable from any ip (ie: 10.0.24.29 or 10.0.14.25 or 10.0.10.5)

_Current behavior_:

keycloak is only reachable through the last ip provided with the `-Djboss.bind.address` (ie: 10.0.10.5)

slaskawi:
Thanks for the feedback @Nowheresly ! That's really helpful!

One of the community members  (Sebastian Schuster) suggested to use `hostname --all-ip-addresses`. Let me try to use that and send another update here.

server/tools/docker-entrypoint.sh
slaskawi: Thanks for the info @belaban! It's similar to what I've seen so far. But it seems @Nowheresly uses a bit more complicated setup.

slaskawi:
@Nowheresly @hmlnarik @stianst I updated the PR and reverted back `CMD` from the Dockerfile. This should allow binding to multiple interfaces. I also followed Sebastian Schuster's advise and used `hostname --all-ip-addresses`.

If you find any free slot, please take a look at it. You may test it out using:
- `docker run slaskawi/keycloak-9767`
- `podman run slaskawi/keycloak-9767`

belaban:
@Nowheresly Perhaps you can use a symbolic name such as `match-address` or `match-host`, or even your own code used by `custom` to determine the bind address to be used by JGroups.
This is described in [1] (scroll down).

[1] http://www.jgroups.org/manual4/index.html#Transport

slaskawi:
@belaban That's a good point Bela! However, I'd like to adjust the bootstrap scripts to work out of the box in such use cases. Hopefully, the updated PR should do the trick (fingers crossed) :)

Nowheresly:
Hi @slaskawi ,
this new version works for me.

Here are the output of some commands:

```
hostname --all-ip-addresses
 10.0.14.48 10.0.10.116 10.0.24.48 172.18.0.8
``` 

```
ps -elf
0 S jboss       272      9 25  80   0 - 604013 futex_ 12:55 ?       00:00:48 /usr/lib/jvm/java/bin/java -D[Standalone] -server -Xms64m -Xmx512m -XX:+UnlockExperimentalVMOptions -XX:+UseCGroupMemoryLimitForHeap -XX:MetaspaceSize=96M -XX:MaxMetaspaceSize=256m -Djava.net.preferIPv4Stack=true -Djboss.modules.system.pkgs=org.jboss.byteman -Djava.awt.headless=true -Dorg.jboss.boot.log.file=/opt/jboss/keycloak/standalone/log/server.log -Dlogging.configuration=file:/opt/jboss/keycloak/standalone/configuration/logging.properties -jar /opt/jboss/keycloak/jboss-modules.jar -mp /opt/jboss/keycloak/modules org.jboss.as.standalone -Djboss.home.dir=/opt/jboss/keycloak -Djboss.server.base.dir=/opt/jboss/keycloak/standalone -Djboss.bind.address=10.0.14.48 -Djboss.bind.address.private=10.0.14.48 -Djboss.bind.address=10.0.10.116 -Djboss.bind.address.private=10.0.10.116 -Djboss.bind.address=10.0.24.48 -Djboss.bind.address.private=10.0.24.48 -Djboss.bind.address=172.18.0.8 -Djboss.bind.address.private=172.18.0.8 -c=standalone-ha.xml -b 0.0.0.0
``` 

I think `-b 0.0.0.0` is mandatory for my use case.

Nowheresly:
Hi @belaban 
Thanks for your reply.
I'm not sure to understand your point.

Our keycloak instance is started in a swarm with a configuration like this one:
docker-compose.yml
```
services:
  keycloak:
    image: jboss/keycloak
    networks:
    - network1
    - network2
    - network3

networks:
  network1:
    driver: overlay
    driver_opts:
      encrypted: "true"
  network2:
    driver: overlay
    driver_opts:
      encrypted: "true"
  network3:
    driver: overlay
    driver_opts:
      encrypted: "true"
```
Now we have 3 different services, each in their own networks, that needs to communicate with keycloak
```
services:
  service1:
    image: service1
    networks:
    - network1
  service2:
    image: service2
    networks:
    - network2
  service3:
    image: service3
    networks:
    - network3
```

All services (ie: `service1`, `service2` or `service3`) can reach keycloak by means of the url : `http://keycloak`. But since each service is in its own network, the internal DNS of docker swarm will resolve the same keycloak instance with different ip for each service.

I'm not sure if my explanation is clear, but I just want to stress out that no service is using directly the ip.

belaban:
@Nowheresly I wanted to stress that there are mechanisms by which JGroups can pick a NIC to bind to. If the solution by @slaskawi  works for you, all the better.
Note that JGroups _cannot_ bind to the wildcard `0.0.0.0` address.

Nowheresly:
ok thank you @belaban for your explanation.

slaskawi:
Thanks for checking this @Nowheresly!

@stianst Could you please retest the new image. If you are happy with it, I believe we can merge this PR.

PR #189, opened 2019-04-17
KEYCLOAK-10082 include docs for exporting Realm from docker container

melissapalmer:
KEYCLOAK-10082 include docs for exporting Realm from docker container


VictorSzewczenko:
I have used these instructions successfully to export a realm, using the `jboss/keycloak:6.0.1`official docker image.

PR #191, opened 2019-04-25
KEYCLOAK-10120: enable statistics with env variable

micedre:
The CLI is run at runtime, when all the other configuration is done (jgroups and database). This allows to have all this configuration in an unique file instead of in all others cli commands.

### DIFF
--- a/server/README.md
+++ b/server/README.md
@@ -421,7 +421,19 @@ containing a `crt` file and point `X509_CA_BUNDLE` environmental variable to tha
 
 NOTE: See `openshift-examples` directory for an out of the box setup for OpenShift.
 
+### Enable some metrics
 
+Keycloak image can collect some statistics for various subsystem which will then be available in the management console and the `/metrics` endpoint.
+You can enable it with the KEYCLOAK_STATISTICS environment variables which take a list of statistics to enable:
+* `db` for the `datasources` subsystem
+* `http` for the `undertow` subsystem
+* `jgroups` for the `jgroups` subsystem
+
+for instance, `KEYCLOAK_STATISTICS=db,http` will enable statistics for the datasources and undertow subsystem.
+
+The special value `all` enables all statistics.
+
+Once enabled, you should see the metrics values changing on the `/metrics` endpoint for the management endpoint.
 
 ## Other details
 
--- a/server/tools/cli/metrics/db.cli
+++ b/server/tools/cli/metrics/db.cli
@@ -1,0 +1,5 @@
+embed-server --server-config=standalone-ha.xml --std-out=echo
+batch
+/subsystem=datasources/data-source=KeycloakDS:write-attribute(name=statistics-enabled, value=true)
+run-batch
+stop-embedded-server
--- a/server/tools/cli/metrics/http.cli
+++ b/server/tools/cli/metrics/http.cli
@@ -1,0 +1,5 @@
+embed-server --server-config=standalone-ha.xml --std-out=echo
+batch
+/subsystem=undertow:write-attribute(name=statistics-enabled,value=true)
+run-batch
+stop-embedded-server
--- a/server/tools/cli/metrics/jgroups.cli
+++ b/server/tools/cli/metrics/jgroups.cli
@@ -1,0 +1,5 @@
+embed-server --server-config=standalone-ha.xml --std-out=echo
+batch
+/subsystem=jgroups/channel=ee:write-attribute(name=statistics-enabled, value=true)
+run-batch
+stop-embedded-server
--- a/server/tools/docker-entrypoint.sh
+++ b/server/tools/docker-entrypoint.sh
@@ -76,6 +76,14 @@ if [ -z "$BIND_OPTS" ]; then
 fi
 SYS_PROPS+=" $BIND_OPTS"
 
+#########################################
+# Expose management console for metrics #
+#########################################
+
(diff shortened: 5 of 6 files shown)


server/README.md
```diff
@@ -315,7 +315,19 @@ containing a `crt` file and point `X509_CA_BUNDLE` environmental variable to tha
 
 NOTE: See `openshift-examples` directory for an out of the box setup for OpenShift.
 
+### Enable some metrics
 
+Keycloak image can collect some statistics for various subsystem which will then be available in the management console and the `/metrics` endpoint.
+You can enable it with the following environment variables :
```
stianst: Should be KEYCLOAK_ not WILDFLY_

EJB3 shouldn't be there as Keycloak doesn't use EJBs.

micedre: sorry about that, I was trying to replicate what was in the server distribution. I didn't realize it was not keycloak specific, i'll fix it.

server/README.md
```diff
@@ -315,7 +315,17 @@ containing a `crt` file and point `X509_CA_BUNDLE` environmental variable to tha
...
+* `KEYCLOAK_DATASOURCES_STATISTICS_ENABLED` for the `datasources` subsystem
+* `KEYCLOAK_UNDERTOW_STATISTICS_ENABLED` for the `undertow` subsystem
+* `KEYCLOAK_JGROUPS_STATISTICS_ENABLED` for the `jgroups` subsystem
+
+You can also enable all of those with `KEYCLOAK_STATISTICS_ENABLED`.
```
stianst: This is currently WILDFLY_STATISTICS_ENABLED in xml file right? So would need to change the value in statistics.cli

micedre: No all was changed with KEYCLOAK prefix following your first comment.

stianst: Never mind. I didn't look properly. I thought there was already a WILDFLY_STATISTICS_ENABLED env variable, but I see you've added it.

stianst:
@micedre thanks for the updates, it looks ready to me now, but would like @slaskawi to give it a test spin to be sure before merging.

slaskawi:
I played a little bit with metrics that we already [1] have. Just as a reminder, you might turn the metrics on by adding `-Dwildfly.datasources.statistics-enabled=true` to the run parameters. If you are interested how such an output looks like, here's an example [3].

Here are some of my thoughts around metrics and this PR:

* The metrics are exposed on the management interface. So I guess we should probably add `-Djboss.bind.address.management=0.0.0.0`, to make the metrics available outside of the container.
* The default XML configuration [2] contains a number of metrics, that can be easily enabled.
* The metrics are prepared to work with a pull-based metrics collection system (something that queries Keycloak metrics endpoint and pulls metrics into a time-series database). Therefore, I'm not sure we should limit what metrics are exposed. In my opinion, we should always expose whatever we've got and let the metrics collection system decide what is needed and what is not.

Having in mind what I wrote the above, I guess the easiest option is to expose all the metrics and maybe create an environment variable that turns Wildfly metrics on (something like `KEYCLOAK_STATISTICS_ENABLED`, that would map to `-Dwildfly.datasources.statistics-enabled=true`).

[1] `docker run -e KEYCLOAK_USER=admin -e KEYCLOAK_PASSWORD=admin jboss/keycloak:6.0.1 -Djboss.bind.address.management=0.0.0.0 -Dwildfly.statistics-enabled=true`
[2] https://gist.github.com/slaskawi/f5721e902b873f75fcd0bce241623d03
[3] https://gist.github.com/slaskawi/5bcb736d8ea20126007afecd3105305a

//cc @vramik

micedre:
`-Dwildfly.datasources.statistics-enabled=true` will not work on the current docker image as the datasource configuration is rewritten at runtime (without the attribute statistics-enabled). This was the reason for this PR.
Now I agree it may be too much to rewrite all flags like that, maybe just adding the missing flags (like on the default zip distribution) would be better ?

stianst:
I prefer the PR as is with KEYCLOAK_ prefixed variables rather than leaking underlying WF container.

Ideally perhaps it would be more usable to have something like:

KEYCLOAK_STATISTICS=none
KEYCLOAK_STATISTICS=all
KEYCLOAK_STATISTICS=db,http

There is some overhead to collecting the stats regardless if they are retrieved or not. I can also see folks not wanting specific stats not wanting the noise, but just have it generated what is needed.

slaskawi:
> -Dwildfly.datasources.statistics-enabled=true will not work on the current docker image as the datasource configuration is rewritten at runtime (without the attribute statistics-enabled). This was the reason for this PR.

@micedre Thanks for clarification. Indeed, missed that bit.

@stianst I like your idea! As long there's one switch to enable all the metrics, I'm happy.

Also, whatever metrics we turn on, it should append `-Djboss.bind.address.management=0.0.0.0` to the `SYS_OPS` I guess.

slaskawi:
@micedre WDYT about @stianst proposal?

micedre:
@slaskawi I like it, but I'm not sure I will have enough time to propose something before september.

micedre:
I have finally a working version, if not super elegant. To add new metrics to enable, simply add a new cli file under the /metrics/ folder. What do you think?

abstractj:
@slaskawi based on Stian's comments and also the fact that @micedre won't have too much time. I'd suggest to us to review the PR as is, future enhancements can be addressed later.

PR #192, opened 2019-04-26
KEYCLOAK-10151 Allow configuration of owner count for the distributed caches

frittentheke:
Issue KEYCLOAK-7582 introduced support for standalone-ha type of clusters via i.e. DNS_PING. Unfortunately the number of owners to the distributed-cache is "1" by default causing data-loss at shutdown of any node. See: https://www.keycloak.org/docs/5.0/server_installation/#_replication

Especially in a container environment like on Kubernetes / OpenShift it just seems natural to have an owner count of 2 or 3 to allow not just for downtime-free, but also data-loss-free rolling updates / reconfiguration of a Keycloak cluster.

Configuration of the owners count was already implemented by in a previous PR by https://github.com/rayscunningham regarding clustering - https://github.com/jboss-dockerfiles/keycloak/pull/96 - but this was closed in favor of https://github.com/jboss-dockerfiles/keycloak/pull/151 later which did not allow such configuration.

### DIFF
--- a/server/README.md
+++ b/server/README.md
@@ -369,6 +369,19 @@ This can be easily achieved by extending the Keycloak image and adding just one 
 
 Of course, we highly encourage you to contribute your custom scripts back to the community image!
 
+### Replication and Fail-Over
+
+By default Keycloak does NOT replicate caches like sessions, authenticationSessions, offlineSessions, loginFailures and a few others (See Eviction and Expiration for more details), which are configured as distributed caches when using a clustered setup. Entries are not replicated to every single node, but instead one or more nodes is chosen as an owner of that data. If a node is not the owner of a specific cache entry it queries the cluster to obtain it. What this means for failover is that if all the nodes that own a piece of data go down, that data is lost forever. By default, Keycloak only specifies one owner for data. So if that one node goes down that data is lost. This usually means that users will be logged out and will have to login again. For more on this subject see [Keycloak Documentation](https://www.keycloak.org/docs/latest/server_installation/#_replication)
+
+#### Specify destributed-cache owners
+
+* `CACHE_OWNERS_COUNT`: Specify the number of distributed-cache owners (default is 1)
+
+AuthenticationSessions will not be replicated by setting CACHE_OWNERS_COUNT>1 as this is usually not required/intended (https://www.keycloak.org/docs/latest/server_installation/#cache)
+To enable replication of AuthenticationSessions as well use:
+
+ * `CACHE_OWNERS_AUTH_SESSIONS_COUNT`: Specify the number of replicas for AuthenticationSessions
+
 
 ## Misc
 
--- a/server/tools/cli/infinispan/cache-owners.cli
+++ b/server/tools/cli/infinispan/cache-owners.cli
@@ -1,0 +1,11 @@
+embed-server --server-config=standalone-ha.xml --std-out=echo
+batch
+/subsystem=infinispan/cache-container=keycloak/distributed-cache=sessions: write-attribute(name=owners, value=${env.CACHE_OWNERS_COUNT:1})
+/subsystem=infinispan/cache-container=keycloak/distributed-cache=offlineSessions: write-attribute(name=owners, value=${env.CACHE_OWNERS_COUNT:1})
+/subsystem=infinispan/cache-container=keycloak/distributed-cache=loginFailures: write-attribute(name=owners, value=${env.CACHE_OWNERS_COUNT:1})
+/subsystem=infinispan/cache-container=keycloak/distributed-cache=clientSessions: write-attribute(name=owners, value=${env.CACHE_OWNERS_COUNT:1})
+/subsystem=infinispan/cache-container=keycloak/distributed-cache=offlineClientSessions: write-attribute(name=owners, value=${env.CACHE_OWNERS_COUNT:1})
+/subsystem=infinispan/cache-container=keycloak/distributed-cache=actionTokens: write-attribute(name=owners, value=${env.CACHE_OWNERS_COUNT:1})
+/subsystem=infinispan/cache-container=keycloak/distributed-cache=authenticationSessions: write-attribute(name=owners, value=${env.CACHE_OWNERS_AUTH_SESSIONS_COUNT:1})
+run-batch
+stop-embedded-server
--- a/server/tools/docker-entrypoint.sh
+++ b/server/tools/docker-entrypoint.sh
@@ -211,6 +211,7 @@ fi
 
 /opt/jboss/tools/x509.sh
 /opt/jboss/tools/jgroups.sh
+/opt/jboss/tools/infinispan.sh
 /opt/jboss/tools/statistics.sh
 /opt/jboss/tools/autorun.sh
 
--- a/server/tools/infinispan.sh
+++ b/server/tools/infinispan.sh
@@ -1,0 +1,14 @@
+# How many owners / replicas should our distributed caches have. If <2 any node that is removed from the cluster will cause a data-loss!
+# As it is only sensible to replicate AuthenticationSessions for certain cases, their replication factor can be configured independently
+
+if [ -n "$CACHE_OWNERS_COUNT" ]; then
+    echo "Setting cache owners to $CACHE_OWNERS_COUNT replicas"
+
+    # Check and log the replication factor of AuthenticationSessions, otherwise this is set to 1 by default
+    if [ -n "$CACHE_OWNERS_AUTH_SESSIONS_COUNT" ]; then
+        echo "Enabling replication of AuthenticationSessions with ${CACHE_OWNERS_AUTH_SESSIONS_COUNT} replicas"
+    else
+        echo "AuthenticationSessions will NOT be replicated, set CACHE_OWNERS_AUTH_SESSIONS_COUNT to configure this"
+    fi
+$JBOSS_HOME/bin/jboss-cli.sh --file="/opt/jboss/tools/cli/infinispan/cache-owners.cli" >& /dev/null
+fi


frittentheke:
@stianst  @slaskawi  ... is there anything I can still contribute to have this reviewed?

stevenscg:
I'm new to keycloak and this type of distributed cache, but I expect that we will get bit by this issue at some point as well. We are using this image with JDBC_PING in a HashiCorp Nomad environment.

frittentheke:
@stevenscg is there anything I can help with or still add to this PR?

stevenscg:
@frittentheke That approach looks good to me. Hoping we can get a review on it from the project team.

I don't know if this is relevant or not, but the `standalone-ha.xml` that I mount into my containers has a few additional named caches that aren't part of your `cache-owners.cli`:

```xml
<distributed-cache name="sessions" owners="1"/>
<distributed-cache name="authenticationSessions" owners="1"/>
<distributed-cache name="offlineSessions" owners="1"/>
<distributed-cache name="clientSessions" owners="1"/>
<distributed-cache name="offlineClientSessions" owners="1"/>
<distributed-cache name="loginFailures" owners="1"/>
<distributed-cache name="actionTokens" owners="2">
```
 
Interesting that the `actionTokens` cache has 2 owners set by default. Should it use your `JGROUPS_CACHE_OWNERS` value as well?

frittentheke:
@stevenscg reading through the documentation at https://www.keycloak.org/docs/6.0/server_installation/index.html#_eviction there is a list of the caches that are of importance in a replicated setup:

`sessions, clientSessions, offlineSessions, offlineClientSessions, loginFailures and actionTokens`

I added the remaining cache containers now and also rebased my branch to the current master.

PTAL

frittentheke:
@slaskawi is there anything I am missing with this PR?

jkroepke:
I would like to see this upstream.

server/tools/jgroups.sh
```diff
@@ -18,3 +18,9 @@ if [ -n "$JGROUPS_DISCOVERY_PROTOCOL" ]; then
        $JBOSS_HOME/bin/jboss-cli.sh --file="/opt/jboss/tools/cli/jgroups/discovery/default.cli" >& /dev/null
     fi
 fi
+
+# How many owners / replicas should our distributed caches have. If <2 any node that is removed from the cluster will cause a data-loss!
+if [ -n "$JGROUPS_CACHE_OWNERS" ]; then
```
slaskawi: This actually should be Infinispan. JGroups project is about cluster communication but it is Infinispan that stores data in a distributed fashion. Perhaps it wouls be even better to name it `NUMBER_OF_CACHE_OWNERS`, @mposolda @stianst @abstractj WDYT?

Apart from that, LGTM.

stianst: It should have a prefix, but JGROUPS is not the correct prefix. Perhaps CACHE_OWNERS_COUNT. One concern I have here is that it is a single value that sets the number of owners for everything. For example authenticationSessions shouldn't need to be replicated as it is a short lived temporary session used during authentication. So at least that cache should be configured separately from the rest.

mposolda: +1 for CACHE_OWNERS_COUNT and authenticationSessions having possibility of separate count. Will be good to document that for authenticationSessions, it is recommended to stick with 1 in most cases. Or maybe even keep this one hardcoded to 1 and document the limitation that authentication eventually needs to start from scratch in case of failover?

slaskawi: > Or maybe even keep this one hardcoded to 1 and document the limitation that authentication eventually needs to start from scratch in case of failover?

@mposolda what about Kubernetes/OpenShift use case where Keycloak operates behind a Load Balancer? You never know which Keycloak instance would be used. Would your proposal work in this scenario?

stianst: It will still work then as it's a distributed cache. Authentication sessions are short lived and updated on every request, so replicating them really isn't good.

slaskawi:
@frittentheke May I ask you to introduce suggested changes?

stianst:
PR is a bit of a mess. Can you please rebase/squash to a single commit?

frittentheke:
@stianst @mposolda @slaskawi  I squashed things into a single commit (again) and used the current master as base.

PTAnotherL

stianst:
@slaskawi can you review and also test this please?

renannprado:
@frittentheke does this number also scale as the number of keycloak replicas scale up in your k8s cluster?

frittentheke:
@renannprado no. 

Actually in my opinion the replication factor can / should be independent from the actual cluster size. The idea is to have potentially "more than one" copy, but not simply replicate everything to every other node (replicas = number of pods).

renannprado:
@frittentheke cool. I just would like to understand if that was intentional. 

Thanks!

server/tools/jgroups.sh
```diff
@@ -27,4 +27,19 @@ if [ -n "$JGROUPS_DISCOVERY_PROTOCOL" ]; then
     else
        $JBOSS_HOME/bin/jboss-cli.sh --file="/opt/jboss/tools/cli/jgroups/discovery/default.cli" >& /dev/null
     fi
+
```
slaskawi: This shouldn't be invoked from `jgroups.sh` file. Perhaps it should be in a separate file - `infinispan.sh`?

frittentheke: Certainly an easy change. But does it make sense to configure replicas without jgroups? I thought the intention was to trigger the clustering that way.

But fair enough, I'll update the PR if we agree it makes sense. What to trigger this new .sh file then?

frittentheke: Reading the documentation about clustering (https://github.com/keycloak/keycloak-containers/blob/master/server/README.md#clustering) again, the jgroup part is only about adjusting how clustering works, but the multicast discovery seems to be enabled by default ... thus the issue https://issues.jboss.org/browse/KEYCLOAK-11009 of not actually being able to "disable" it. So externalizing the configuration snippets for replicasd out of the jgroups configuration makes total sense now.

server/tools/cli/jgroups/cache-owners.cli
```diff
@@ -0,0 +1,7 @@
+/subsystem=infinispan/cache-container=keycloak/distributed-cache=sessions: write-attribute(name=owners, value=${env.CACHE_OWNERS_COUNT:1})
```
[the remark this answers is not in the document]
frittentheke: @slaskawi your test failed because you did not set JGROUPS_DISCOVERY_PROTOCOL as was required to even consider CACHE_OWNERS_COUNT / CACHE_OWNERS_AUTH_SESSIONS_COUNT.

As discussed I changed that just now and force pushed a singe commit to this branch again.
I also built and tested the Docker image myself, seems to work as you expect now.

PTAnotherL

frittentheke:
@slaskawi @mposolda PTAnotherL

server/tools/jgroups.sh
slaskawi: Yes, this is exactly the point. Default discovery protocols might be good enough for some cases.

server/tools/cli/jgroups/cache-owners.cli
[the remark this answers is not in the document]
slaskawi: Ok, will do it now.

server/tools/jgroups.sh
```diff
@@ -27,4 +27,19 @@ if [ -n "$JGROUPS_DISCOVERY_PROTOCOL" ]; then
...
+   # How many owners / replicas should our distributed caches have. If <2 any node that is removed from the cluster will cause a data-loss!
+   # As it is only sensible to replicate AuthenticationSessions for certain cases, their replication factor can be configured independently
+   if [ -n "$CACHE_OWNERS_COUNT" ]; then
```
slaskawi: I believe you left this modification by a mistake? All the changes seems to be implemented in `infinispan.sh`. If I'm right, could you please update this PR once again and remove this part?

frittentheke: @slaskawi  .. argh sorry about that. Fixed it.

frittentheke:
@slaskawi added that and now the resulting file /opt/jboss/keycloak/standalone/configuration/standalone-ha.xml contains the settings.

frittentheke:
@mposolda PTAL

slaskawi:
@frittentheke Ok, so it seems to work. 

I tested the container using the following command: `docker run -e CACHE_OWNERS_COUNT=3 -e CACHE_OWNERS_AUTH_SESSIONS_COUNT=2 slaskawi-keycloak:PR-192-3` and this renders the following configuration:

```
<distributed-cache name="sessions" owners="${env.CACHE_OWNERS_COUNT:1}"/>
<distributed-cache name="authenticationSessions" owners="${env.CACHE_OWNERS_AUTH_SESSIONS_COUNT:1}"/>
<distributed-cache name="offlineSessions" owners="${env.CACHE_OWNERS_COUNT:1}"/>
<distributed-cache name="clientSessions" owners="${env.CACHE_OWNERS_COUNT:1}"/>
<distributed-cache name="offlineClientSessions" owners="${env.CACHE_OWNERS_COUNT:1}"/>
<distributed-cache name="loginFailures" owners="${env.CACHE_OWNERS_COUNT:1}"/>
<distributed-cache name="actionTokens" owners="${env.CACHE_OWNERS_COUNT:1}">
```

JBoss CLI produces proper output (note, that you need to `resolve-expressions=true`):

```
/subsystem=infinispan/cache-container=keycloak/distributed-cache=sessions:read-resource(resolve-expressions=true)
{
    "outcome" => "success",
    "result" => {
        "async-marshalling" => false,
        "batching" => false,
        "capacity-factor" => 1.0,
        "consistent-hash-strategy" => "INTER_CACHE",
        "indexing" => "NONE",
        "indexing-properties" => undefined,
        "jndi-name" => undefined,
        "l1-lifespan" => 0L,
        "mode" => "SYNC",
        "module" => undefined,
        "owners" => "3",
        "queue-flush-interval" => 10L,
        "queue-size" => 0,
        "remote-timeout" => 10000L,
        "segments" => 256,
        "start" => "LAZY",
        "statistics-enabled" => false,
        "backup-for" => {"BACKUP_FOR" => undefined},
        "component" => {
            "backup-for" => undefined,
            "backups" => undefined,
            "expiration" => undefined,
            "locking" => undefined,
            "partition-handling" => undefined,
            "state-transfer" => undefined,
            "transaction" => undefined,
            "eviction" => undefined
        },
        "eviction" => {"EVICTION" => undefined},
        "expiration" => {"EXPIRATION" => undefined},
        "locking" => {"LOCKING" => undefined},
        "memory" => {"object" => undefined},
        "state-transfer" => {"STATE_TRANSFER" => undefined},
        "store" => {"none" => undefined},
        "transaction" => {"TRANSACTION" => undefined}
    }
}
```

So far, most of our scripts were rendering exact values in the configuration. This one puts references to the environment (like `${env.CACHE_OWNERS_COUNT:1}`). I'm not against it, just wanted to confirm if this is something you wanted to achieve, @frittentheke ? The biggest downside as you probably noticed is that you need to use the `jboss-cli.sh` to verify that proper values are resolved and injected where you wanted (e.g. that there is no typo there). 

However, if this does work for you, I'm OK with that. @stianst @abstractj @mposolda Any comments from your side?

PR #194, opened 2019-05-03
KEYCLOAK-10189 Possibility to use Gitlab for build keycloak sources

Kilian22:
I add the possibility to use for exemple Gitlab repo when you want to build a new keycloak image from a non official repository.

### DIFF
--- a/server/Dockerfile
+++ b/server/Dockerfile
@@ -12,6 +12,7 @@ ENV LANG en_US.UTF-8
 
 ARG GIT_REPO
 ARG GIT_BRANCH
+ARG GIT_HOST
 ARG KEYCLOAK_DIST=https://downloads.jboss.org/keycloak/$KEYCLOAK_VERSION/keycloak-$KEYCLOAK_VERSION.tar.gz
 
 USER root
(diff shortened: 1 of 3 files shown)


slaskawi:
Even though I like this change, but we keep Keycloak sources in Github in general. So I'm not sure if this is something that should belong to a custom user repo rather than this one. 

But let's see what @stianst thinks about it.

server/Dockerfile
```diff
@@ -12,6 +12,7 @@ ENV LANG en_US.UTF-8
 
 ARG GIT_REPO
 ARG GIT_BRANCH
+ARG GIT_HOST
```
stianst: Set default value to https://github.com here rather than in build-keycloak.sh.

server/README.md
```diff
@@ -335,6 +335,12 @@ It is possible to build Keycloak from a GitHub repository instead of downloading
 
 This will clone the repository then build Keycloak from source. If you don't include GIT_BRANCH it will use the `master` branch.
 
+You can also use another solution of repository like Gitlab using GIT_HOST option.
```
stianst: No need to document this. I really don't think anyone else will use it.

server/tools/build-keycloak.sh
```diff
@@ -9,20 +9,25 @@ if [ "$GIT_REPO" != "" ]; then
         GIT_BRANCH="master"
     fi
 
+if [ "$GIT_HOST" == "" ]; then
```
stianst: Remove, use default value for build arg as mentioned above

stianst:

PR #195, opened 2019-05-14
Update to the latest KeyCloak version

zagaria:


stianst:
Thanks, we will update our release jobs to do this automatically in the future

PR #201, opened 2019-06-13
KEYCLOAK-10652 Truststore optimizations

slaskawi:
https://issues.jboss.org/browse/KEYCLOAK-10652

This Pull Request allows adding multiple CA Bundles into `X509_CA_BUNDLE` and imports generated trust store into Truststore SPI. Here's an example:

```bash
X509_CA_BUNDLE=/var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt /var/run/secrets/kubernetes.io/serviceaccount/ca.crt
```

When running in Kubernetes with the example above, Keycloak will automatically be able to call other services deployed in OpenShift (using `service-ca.crt`) as we as the Kubernetes API server (using `ca.crt`).

### DIFF
--- a/server/README.md
+++ b/server/README.md
@@ -416,8 +416,7 @@ Keycloak image allows you to specify both a private key and a certificate for se
 
 Those files need to be mounted in `/etc/x509/https` directory. The image will automatically convert them into a Java keystore and reconfigure Wildfly to use it.
 
-It is also possible to provide an additional CA bundle and setup Mutual TLS this way. In that case, you need to mount an additional volume to the image
-containing a `crt` file and point `X509_CA_BUNDLE` environmental variable to that file.
+It is also possible to provide an additional CA bundle and setup Mutual TLS this way. In that case, you need to mount an additional volume (or multiple volumes) to the image. These volumes should contain all necessary `crt` files. The final step is to configure the `X509_CA_BUNDLE` environment variable to contain a list of the locations of the various CA certificate bundle files specified before, separated by space (` `). In case of an OpenShift environment, that could be `/var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt /var/run/secrets/kubernetes.io/serviceaccount/ca.crt`.
 
 NOTE: See `openshift-examples` directory for an out of the box setup for OpenShift.
 
--- a/openshift-examples/keycloak-https-mutual-tls.json
+++ b/openshift-examples/keycloak-https-mutual-tls.json
@@ -236,7 +236,7 @@
                                     },
                                     {
                                         "name": "X509_CA_BUNDLE",
-                                        "value": "/var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt"
+                                        "value": "/var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt /var/run/secrets/kubernetes.io/serviceaccount/ca.crt"
                                     },
                                     {
                                         "name": "JGROUPS_DISCOVERY_PROTOCOL",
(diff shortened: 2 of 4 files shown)


stianst:
Should probably also update readme. Also, not sure if multiple truststores can always be expected to have some password, etc.?

stianst:
Please also provide a description in the JIRA. Why is this needed? I can probably second guess it, but shouldn't have to. Not sure the summary is all that meaningful either.

slaskawi:
@stianst This is still work-in-progress. I'll fill all the details once everything works fine (I'm testing it up together with Eclipse Che Team).

slaskawi:
> Should probably also update readme. Also, not sure if multiple truststores can always be expected to have some password, etc.?

I added a readme entry.

Yes, that's absolutely fine. Since Java can not read CRT files out of the box, we need to pick them up and transform into JKS file. This happens automatically in `x509.sh` file.

> Please also provide a description in the JIRA. Why is this needed? I can probably second guess it, but shouldn't have to. Not sure the summary is all that meaningful either.

The JIRA will be updated in a minute. Thanks for the reminder.

slaskawi:
@davidfestal The change can be tested by using this Docker image: `slaskawi/keycloak-openshift-4`

nokernel:
:+1:  on this is needed here

slaskawi:
@abstractj @iankko Could you please take a look at this one?

abstractj:
@slaskawi I don't see the readme entry in these changes. Where should I look?

slaskawi:
@abstractj that's a good point, let me add them.

slaskawi:
@abstractj I added all necessary information.

server/README.md
```diff
@@ -381,8 +381,8 @@ Keycloak image allows you to specify both a private key and a certificate for se
 
 Those files need to be mounted in `/etc/x509/https` directory. The image will automatically convert them into a Java keystore and reconfigure Wildfly to use it.
 
-It is also possible to provide an additional CA bundle and setup Mutual TLS this way. In that case, you need to mount an additional volume to the image
-containing a `crt` file and point `X509_CA_BUNDLE` environmental variable to that file.
+It is also possible to provide an additional CA bundle and setup Mutual TLS this way. In that case, you need to mount an additional volume (or multiple volumes) to the image the contains all necessary `crt` files. The final step is to point `X509_CA_BUNDLE`
```
iankko: @slaskawi ```s/... to the image the contains all necessary `crt` files/... to the image. These volumes should contain all necessary .../g```

server/README.md
```diff
@@ -381,8 +381,8 @@ Keycloak image allows you to specify both a private key and a certificate for se
...
+ environmental variable to them using space (` `) a separator. In case of an OpenShift environment, that could be `/var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt /var/run/secrets/kubernetes.io/serviceaccount/ca.crt`.
```
iankko: ```s/The final step is to point `X509_CA_BUNDLE` environmental variable to them using space (` `) a separator./The final step is to configure the `X509_CA_BUNDLE` environment variable to contain a list of the locations of the various CA certificate bundle files specified before, separated by space (` `)./g```

```s/In case of an OpenShift environment, that could be ..../For example, in the case of an OpenShift environment, the `X509_CA_BUNDLE` value would be .../g```

server/README.md
slaskawi: Fixed thanks!

server/README.md
slaskawi: Fixed thanks!

slaskawi:
> Just 2-3 small nitpicks to rephrase certain parts of the README.md description. The rest lgtm

@iankko Thanks for the review. All comments addressed.

PR #203, opened 2019-07-05
KEYCLOAK-10789 Add optional unparsed JGROUPS_DISCOVERY_PROPERTIES_DIRECT

YenTheFirst:
The parsing step of JGROUPS_DISCOVERY_PROPERTIES in jgroups.sh makes
it impossible to correctly specify properties whose value includes an
`=` character. Some discovery protocols, such as KUBE_PING, require
properties that have an `=` in their value, i.e., `{labels=>app=keycloak}`

### DIFF
--- a/server/README.md
+++ b/server/README.md
@@ -271,12 +271,16 @@ all scripts shipped in the image.
 ## Clustering
 
 Replacing the default discovery protocols (`PING` for the UDP stack and `MPING` for the TCP one) can be achieved by defining
-two additional environment variables:
+some additional environment variables:
 
 - `JGROUPS_DISCOVERY_PROTOCOL` - name of the discovery protocol, e.g. DNS_PING
 - `JGROUPS_DISCOVERY_PROPERTIES` - an optional parameter with the discovery protocol properties in the following format:
 `PROP1=FOO,PROP2=BAR`
+- `JGROUPS_DISCOVERY_PROPERTIES_DIRECT` - an optional parameter with the discovery protocol properties in jboss CLI format:
+`{PROP1=>FOO,PROP2=>BAR}`
 - `JGROUPS_TRANSPORT_STACK` - an optional name of the transport stack to use `udp` or `tcp` are possible values. Default: `tcp` 
+
+**Warning**: It's an error to set both JGROUPS_DISCOVERY_PROPERTIES and JGROUPS_DISCOVERY_PROPERTIES_DIRECT. No more than one of them may be set.
 
 The bootstrap script will detect the variables and adjust the `standalone-ha.xml` configuration based on them.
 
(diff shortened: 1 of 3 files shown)


slaskawi:
Thanks a lot for the Pull Request!

Indeed, it seems the current implementation has some problems while parsing `=` sign. However, I would prefer to fix the parsing script (probably replace the `sed` with something better) and not change the format of the `JGROUPS_DISCOVERY_PROPERTIES` property. 

Please note that if we changed it, some application may blow up during the upgrade (because we don't provide any backwards compatibility). 

@YenTheFirst Could you please reconsider this PR and possibly upload a new PR with fixed parsing script?

YenTheFirst:
What would you propose is a reasonable format for
JGROUPS_DISCOVERY_PROPERTIES when some of the values contain `=`?
`key1=value1,key2=foo=bar` ?

This also doesn't address what we'd need to do about values that wanted to
contain a comma, though I'm not aware off the top of my head of any
properties of any protocols that need a comma.

slaskawi:
> What would you propose is a reasonable format for JGROUPS_DISCOVERY_PROPERTIES when some of the values contain `=`? `key1=value1,key2=foo=bar` ?

@YenTheFirst That's a good point. After thinking about this for a while, two solutions came into my mind:

1. We might implement some sort of escaping. E.g. `key1=value1,key2=foo=bar` could become `key1=value1,key2=foo\=bar`. That's a very common approach in similar situations.
2. We might introduce the new format but also scan the first character. If it's `{` (so the new format), we might assign obtained string directly to `JGROUPS_DISCOVERY_PROPERTIES_PARSED`. If it's something other than '{' (so the old format) we might proceed with the parsing step. 

The biggest advantage of no. 2 is that it's backwards compatible (so we don't even have to change the OpenShift templates) and it introduces a new format, just in case you need `=`.

The only problem I can foresee is dealing with JSON-formatted properties. In other words, how to interpret `JGROUPS_DISCOVERY_PROPERTIES={"key": "val"}`. I'm not sure how the CLI would react to that. But on the other hand, that's an uncommon use case.

YenTheFirst:
a third option that comes to mind is

3. take a new parameter, JGROUPS_DISCOVERY_PROPERTIES_RAW or similar, which
takes the raw jboss format, and it's an error to pass more than one of the
environment variables

slaskawi:
@YenTheFirst Yes, no. 3 is also a valid approach. The only drawback I see is that having two properties effectively setting the same thing is always error prone. But as you said, we can catch such a situation and print out an error message.

Personally, I'm slightly leaning towards option no. 2 (dynamically checking what is the format of the property and adjusting parsing to it). But I'm also OK with option no. 3. Let me leave the final choice to you.

YenTheFirst:
I've updated the PR to use option no. 3, adding a new configuration variable, `JGROUPS_DISCOVERY_PROPERTIES_DIRECT`.

This should allow any existing deployments to continue working as-is, while allowing users to explicitly opt-in to the non-parsing behavior.

slaskawi:
@abstractj This one is ready to be integrated.

PR #210, opened 2019-08-15
KEYCLOAK-11124 add dns. to example JGroup protocol

max-allan-surevine:
KEYCLOAK-11124
Adding dns. makes the example JGROUPS_DISCOVERY_PROTOCOL not fail.


slaskawi:
Thank you for spotting this @max-allan-surevine!

@abstractj This one is ready to be integrated.

PR #212, opened 2019-08-23
KEYCLOAK-11182 delegate jdbc jar download to mvn to allow better control of dependency distro

greghall76:
allows mirrors for source or dependency download
https://issues.jboss.org/browse/KEYCLOAK-11182


server/tools/build-keycloak.sh
```diff
@@ -16,13 +16,13 @@ if [ "$GIT_REPO" != "" ]; then
     export M2_HOME=/opt/jboss/maven
 
     # Clone repository
-    git clone --depth 1 https://github.com/$GIT_REPO.git -b $GIT_BRANCH /opt/jboss/keycloak-source
+    git clone --depth 1 $GIT_REPO -b $GIT_BRANCH /opt/jboss/keycloak-source
```
slaskawi: @stianst This one can blow up our CI jobs. May I ask you for opinion here?

server/tools/build-keycloak.sh
```diff
@@ -50,17 +50,29 @@ fi
...
+else
+  curl -L https://repo1.maven.org/maven2/org/postgresql/postgresql/$JDBC_POSTGRES_VERSION/postgresql-$JDBC_POSTGRES_VERSION.jar > postgres-jdbc.jar
+fi
 cp /opt/jboss/tools/databases/postgres/module.xml .
 
 mkdir -p /opt/jboss/keycloak/modules/system/layers/base/org/mariadb/jdbc/main
 cd /opt/jboss/keycloak/modules/system/layers/base/org/mariadb/jdbc/main
-curl -L https://repo1.maven.org/maven2/org/mariadb/jdbc/mariadb-java-client/$JDBC_MARIADB_VERSION/mariadb-java-client-$JDBC_MARIADB_VERSION.jar > mariadb-jdbc.jar
+if [ -x "$(command -v mvn)" ]; then
+  mvn dependency:get -Dartifact=org/mariadb/jdbc:mariadb-java-client:$JDBC_MARIADB_VERSION:jar -Ddest=mariadb-jdbc.jar
```
slaskawi: Personally, I'm not a big fan of this approach. 

It adds additional complexity by introducing `if/else` blocks and it doesn't really buy us anything. It also increases overall build time as Maven checks also `maven-metadata.xml`.

For my personal taste, using `curl` and hardcoding Maven Central repo here is enough. But maybe others (@abstractj @stianst @iankko) like this approach.

stianst: I'm against these changes as well. To solve the use-cases in this case I'd be happy to have the Maven repo set as a variable like we did for download repo.

stianst:
Closing this PR. It's sent against the wrong branch. Need to be against master branch. Also, the approach is not great. It would be better to use a MAVEN_REPO arg.

PR #218, opened 2019-09-18
KEYCLOAK-11419 Vault support in Docker

martin-kanis:
This PR requires Keycloak version 8.0.0 where the vault support is present.

### DIFF
--- a/server/tools/vault.sh
+++ b/server/tools/vault.sh
@@ -1,0 +1,13 @@
+#!/bin/bash
+
+if [ -d "$JBOSS_HOME/secrets" ]; then
+    echo "set plaintext_vault_provider_dir=${JBOSS_HOME}/secrets" >> "$JBOSS_HOME/bin/.jbossclirc"
+
+    echo "set configuration_file=standalone.xml" >> "$JBOSS_HOME/bin/.jbossclirc"
+    $JBOSS_HOME/bin/jboss-cli.sh --file=/opt/jboss/tools/cli/files-plaintext-vault.cli
+    sed -i '$ d' "$JBOSS_HOME/bin/.jbossclirc"
+
+    echo "set configuration_file=standalone-ha.xml" >> "$JBOSS_HOME/bin/.jbossclirc"
+    $JBOSS_HOME/bin/jboss-cli.sh --file=/opt/jboss/tools/cli/files-plaintext-vault.cli
+    sed -i '$ d' "$JBOSS_HOME/bin/.jbossclirc"
+fi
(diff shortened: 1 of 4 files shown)


stianst:
Mounting a directory on plain Docker doesn't make much sense, so not sure we should document it like this. Makes more sense to provide some documentation around Kubernetes and OpenShift. I'm also still not liking the whole PLAINTEXT thing.

Would it not be better to just have the provider enabled by default and pointing to $KEYCLOAK_HOME/secrets. Then instructions would be simpler as we'd have instructions for how to mount secrets from Kubernetes into secrets dir?

slaskawi:
@stianst @hmlnarik @martin-kanis So what do we plan to do with this PR?

stianst:
My previous comment still applies. Shouldn't need to enable it and it should just pick up secrets from KEYCLOAK_HOME/secrets without any extra config. Docs should be updated to describe where secrets are mounted, but not give instructions with docker. Should also mention that this is useful to mount secrets from Kubernetes and Openshift

martin-kanis:
@slaskawi @stianst I plan to address Stian's comments when I'm done with other priorities. (7.3.4, WF-18 upgrade)

slaskawi:
@martin-kanis @stianst I would suggest to close this PR than and re-open it once it's ready for review.

stianst:
I see no issue with leaving it open as I assume the changes will be done in the next few weeks. If it was months from now I would agree closing the PR would be best.

server/tools/vault.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+if [ -d "$JBOSS_HOME/secrets" ]; then
+    echo "set plaintext_vault_provider_dir=${JBOSS_HOME}/secrets" >> "$JBOSS_HOME/bin/.jbossclirc"
```
slaskawi: You probably should clear `$JBOSS_HOME/bin/.jbossclirc` before appending it. But that's a minor thing it should work fine even without it.

ChristianCiach:
Please excuse me for replying to such an old PR, but I would like to see some clarification.

The README currently still states:

> ### Setup Kubernetes / OpenShift files plaintext vault
> 
> Keycloak supports vault implementation for [Kubernetes secrets](https://kubernetes.io/docs/concepts/configuration/secret/). To use files plaintext vault in Docker container mount secret files to `$JBOSS_HOME/secrets` directory. This can be used to consume secrets from Kubernetes / OpenShift cluster. 

Excuse me, but why are you even talking about? There is no such thing as a "Kubernetes vault" or a "Docker vault". Could you please clarify, for future reference, what you mean by "vault"? Like Ansible vault?

PR #220, opened 2019-09-30
KEYCLOAK-11590 Add unprivileged user

oyelekci:
Tested building the image locally.


slaskawi:
@abstractj Could you please have a look at it?

PR #224, opened 2019-10-09
[KEYCLOAK-11688] MSSQL instance name

arminfelder:
add ENV variable DB_INSTANCE_NAME for MSSQL, to configure the instance name for Microsoft SQLServer

### DIFF
--- a/server/tools/cli/databases/mssql/change-database.cli
+++ b/server/tools/cli/databases/mssql/change-database.cli
@@ -1,5 +1,5 @@
 /subsystem=datasources/data-source=KeycloakDS: remove()
-/subsystem=datasources/data-source=KeycloakDS: add(jndi-name=java:jboss/datasources/KeycloakDS,enabled=true,use-java-context=true,use-ccm=true, connection-url="jdbc:sqlserver://${env.DB_ADDR:mssql}:${env.DB_PORT:1433};databaseName=${env.DB_DATABASE:keycloak};sendStringParametersAsUnicode=false;integratedSecurity=false;user=${env.DB_USER:keycloak};password=${env.DB_PASSWORD:password};${env.JDBC_PARAMS:}", driver-name=sqlserver)
+/subsystem=datasources/data-source=KeycloakDS: add(jndi-name=java:jboss/datasources/KeycloakDS,enabled=true,use-java-context=true,use-ccm=true, connection-url="jdbc:sqlserver://${env.DB_ADDR:mssql}:${env.DB_PORT:1433};instanceName=${env.DB_INSTANCE_NAME};databaseName=${env.DB_DATABASE:keycloak};sendStringParametersAsUnicode=false;integratedSecurity=false;user=${env.DB_USER:keycloak};password=${env.DB_PASSWORD:password};${env.JDBC_PARAMS:}", driver-name=sqlserver)
 /subsystem=datasources/data-source=KeycloakDS: write-attribute(name=user-name, value=${env.DB_USER:keycloak})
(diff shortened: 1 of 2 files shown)


stianst:
I'm not an expert on MSSQL by any means, but looking at https://social.msdn.microsoft.com/Forums/sqlserver/en-US/eb58a7fe-4647-4eea-91e3-6ee369362d64/how-to-find-instance-name?forum=sqlgetstarted it seems that instanceName is empty by default, so this change would break things. Seems instanceName should only be added to connection string if specified.

arminfelder:
> I'm not an expert on MSSQL by any means, but looking at https://social.msdn.microsoft.com/Forums/sqlserver/en-US/eb58a7fe-4647-4eea-91e3-6ee369362d64/how-to-find-instance-name?forum=sqlgetstarted it seems that instanceName is empty by default, so this change would break things. Seems instanceName should only be added to connection string if specified.

I am not an MS SQL expert either, but according to https://docs.microsoft.com/bs-cyrl-ba/sql/sql-server/install/instance-configuration?view=sql-server-2014 MSSQLSERVER should be default

slaskawi:
@hmlnarik @vramik @mhajas @vmuzikar Perhaps you guys have more knowledge on this?

vmuzikar:
@slaskawi Sorry, I know nothing about this. :)

ibauersachs:
Is this really necessary at all? Usually the instance name is just specified as `server\instance`.

The existing connection string in the changed file probably won't with named instances anyway. They usually use dynamic ports which use the [SQL Server Browser service](https://docs.microsoft.com/en-us/sql/tools/configuration-manager/sql-server-browser-service) on port 1434 to find the actual port. Explicitly specifying 1433 (the default port for the default instance) overrides this.

slaskawi:
@arminfelder Would you mind to answer @ibauersachs 's comment?

arminfelder:
@ibauersachs I don't know much about sql server browser service, but I do have this code in production with several MSSQL Datacenter Edition instances, with different instancenames on each one

slaskawi:
To me that explanation seems legit.

ibauersachs:
@arminfelder On which ports are your datacenter instances running? If I read https://docs.microsoft.com/en-us/sql/connect/jdbc/building-the-connection-url?view=sql-server-ver15 correctly, the instance name is probably irrelevant in your case.

ibauersachs:
So, a test on my local machine that has two Named Instances installed (Developer Editions 2017 and 2019) fails with the current docker image because of the default port:

```
...
Caused by: com.microsoft.sqlserver.jdbc.SQLServerException:
  The TCP/IP connection to the host host.docker.internal, port 1433 has failed.
....
```

If I specify the dynamic port, the connection succeeds. But then I don't need to specify an instance name, which matches the documentation:
> For optimal connection performance, you should set the portNumber when you connect to a named instance. This will avoid a round trip to the server to determine the port number. If both a portNumber and instanceName are used, the portNumber will take precedence and the instanceName will be ignored.

So my conclusion is:
- Remove the default port
- This PR is not necessary, the instance name can already be set by specifying `server\instance`, the `instanceName=` is an additional JDBC property not found in other MSSQL connection strings. It overrides the instance if it is already specified in the server name (not 100% sure, guessing from reading the [source](https://github.com/microsoft/mssql-jdbc/blob/v8.3.1/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java#L1486)). The PR probably doesn't break existing installations though because of the default port issue.

slaskawi:
After some lengthy discussions, I'm leaning towards @ibauersachs 's explanation that this PR is not necessary. 

@hmlnarik @douglaspalmer @abstractj @stianst Please reopen if you don't agree with me.

PR #221, opened 2019-10-02
[KEYCLOAK-11622] Make the user/password handling more safe for whitespace

cartmanume:
Make the user/password handling more safe for whitespace in user or password. This wraps them both in ".

PR #225, opened 2019-10-10
[KEYCLOAK-11696] Improvements of docker-entrypoint.sh

cartmanume:
I ran https://github.com/koalaman/shellcheck to see if it could be improved and found some minor stuff that would make if more safe. Also added `-e` to make it fail if any subcommand fails.
More details in the JIRA ticket.
https://issues.jboss.org/browse/KEYCLOAK-11696

### DIFF
--- a/server/tools/docker-entrypoint.sh
+++ b/server/tools/docker-entrypoint.sh
@@ -1,4 +1,5 @@
-#!/bin/bash -e
+#!/bin/bash
+set -eou pipefail
 
 # usage: file_env VAR [DEFAULT]
 #    ie: file_env 'XYZ_DB_PASSWORD' 'example'
--- a/server/tools/docker-entrypoint.sh
+++ b/server/tools/docker-entrypoint.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/bash -e
 
 # usage: file_env VAR [DEFAULT]
 #    ie: file_env 'XYZ_DB_PASSWORD' 'example'
@@ -93,7 +93,7 @@ fi
 #################
 
 # If the server configuration parameter is not present, append the HA profile.
-if echo "$@" | egrep -v -- '-c |-c=|--server-config |--server-config='; then
+if echo "$@" | grep -E -v -- '-c |-c=|--server-config |--server-config='; then
     SYS_PROPS+=" -c=standalone-ha.xml"
 fi
 
@@ -105,7 +105,7 @@ file_env 'DB_USER'
 file_env 'DB_PASSWORD'
 
 # Lower case DB_VENDOR
-DB_VENDOR=`echo $DB_VENDOR | tr A-Z a-z`
+DB_VENDOR=$(echo "$DB_VENDOR" | tr "[:upper:]" "[:lower:]")
 
 # Detect DB vendor from default host names
 if [ "$DB_VENDOR" == "" ]; then
@@ -181,20 +181,21 @@ case "$DB_VENDOR" in
 esac
 
 # Append '?' in the beggining of the string if JDBC_PARAMS value isn't empty
-export JDBC_PARAMS=$(echo ${JDBC_PARAMS} | sed '/^$/! s/^/?/')
+JDBC_PARAMS=$(echo "${JDBC_PARAMS}" | sed '/^$/! s/^/?/')
+export JDBC_PARAMS
 
 # Convert deprecated DB specific variables
 function set_legacy_vars() {
   local suffixes=(ADDR DATABASE USER PASSWORD PORT)
   for suffix in "${suffixes[@]}"; do
     local varname="$1_$suffix"
-    if [ ${!varname} ]; then
-      echo WARNING: $varname variable name is DEPRECATED replace with DB_$suffix
-      export DB_$suffix=${!varname}
+    if [ "${!varname}" ]; then
+      echo WARNING: "$varname" variable name is DEPRECATED replace with DB_"$suffix"
+      export DB_"$suffix=${!varname}"
     fi
   done
 }
-set_legacy_vars `echo $DB_VENDOR | tr a-z A-Z`
+set_legacy_vars "$(echo "$DB_VENDOR" | tr "[:upper:]" "[:lower:]")"
 
 # Configure DB
 


server/tools/docker-entrypoint.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/bash -e
```
SuperSandro2000: why not ``set -e`` in a separate line?
Also we probably want ``set -eou pipefail``

cartmanume: Yes, that would also work. About the other options I am not sure it that would break, I did not test with them. But I agree that it would be very nice to have them.

SuperSandro2000: `-o pipefail` would only forward errors trough pipes so if the first command fails the entire line fails. Very good combo with `-e`.
`-u` errors when unset variables are accessed. This normally needs some `${VAR:-}` to fall back to empty and not error.

slaskawi:
Personally, I'm not in favor of adding fixes noticed by static analysis tools and not introducing any functional changes nor fixing bugs. We had a similar situation in the past, where potentially trivial change implemented to make a tool happy broke clustering pretty badly. So I have my reasons to be skeptical here.

I would just ask the most obvious thing - what tests did you perform to make sure everything works correctly?

Maybe some others (@stianst , @abstractj ?) will be more in favor of accepting this change.

cartmanume:
> Personally, I'm not in favor of adding fixes noticed by static analysis tools and not introducing any functional changes nor fixing bugs. We had a similar situation in the past, where potentially trivial change implemented to make a tool happy broke clustering pretty badly. So I have my reasons to be skeptical here.
> 
> I would just ask the most obvious thing - what tests did you perform to make sure everything works correctly?
> 
> Maybe some others (@stianst , @abstractj ?) will be more in favor of accepting this change.

Totally understandable. What I did to test was to build and start the container in docker and see that it started without any errors in the output.

server/tools/docker-entrypoint.sh
cartmanume: I know what they do and I gave them both a thought, but I also wanted to minimize the changes of how the script behaves. Adding `-u` would need a lot more extensive testing of it for example.

abstractj:
@cartmanume thanks for contributing to this, one thing that's confusing to me about this change is the fact that I'm not sure what we're aiming to fix. 
I totally understand that tools like `shellcheck` can be helpful to suggest enhancements, but at the same time I wonder if you're having any issues which the current script? If yes, do you have the steps to reproduce? Despite the suggestion from `shellcheck`, why we need this change?

cartmanume:
> @cartmanume thanks for contributing to this, one thing that's confusing to me about this change is the fact that I'm not sure what we're aiming to fix.
> I understand that tools like `shellcheck` can be helpful to suggest enhancements, but at the same time, I wonder if you're having any issues which the current script? If yes, do you have the steps to reproduce? Despite the suggestion from `shellcheck`, why we need this change?

Now when looking at this after a while I think that it should have been two separate requests.

The `-e` we had real issues with in production, we wanted a password containing spaces and #221 fixed that, but the line itself failed and printed an error in the startup log and the password was not set at all. With `-e` the container wouldn't have started and the reason why the password was not set would be visible directly and forced to react on.

The other parts are not all that was found by `shellcheck`, I picked out the ones that I recognized and that we have had real issues in in other projects at my work, for example not quoting variables can lead to confusing problems, as in #221.

I can also understand the approach to not fix something that aint broken. But in my opinion this is covered under code quality. I guess if someone used `==` to compare strings in Java and it works for now since the strings are specified in compile time you would probably want to change that to use `equals` instead, right?

SuperSandro2000:
It is considered best practice to use ``set -eou pipefail``. You could read about the topic: http://redsymbol.net/articles/unofficial-bash-strict-mode/

abstractj:
> Now when looking at this after a while I think that it should have been two separate requests.
> 
> The `-e` we had real issues with in production, we wanted a password containing spaces and #221 fixed that, but the line itself failed and printed an error in the startup log and the password was not set at all. With `-e` the container wouldn't have started and the reason why the password was not set would be visible directly and forced to react on.
> 
> The other parts are not all that was found by `shellcheck`, I picked out the ones that I recognized and that we have had real issues in in other projects at my work, for example not quoting variables can lead to confusing problems, as in #221.

Sure np, just trying to understand. Do you still have issues when docker-entrypoint.sh starts? Do you have any scenario where the script fails but still, the container starts?

> 
> I can also understand the approach to not fix something that aint broken. But in my opinion this is covered under code quality. I guess if someone used `==` to compare strings in Java and it works for now since the strings are specified in compile time you would probably want to change that to use `equals` instead, right?

Yes and no, there are several static analysis tools out there. But not everything that's suggested, should be applied to the codebase. Tools and best practices are here to help but in the end, we need to use our best judgment.

@iankko @douglaspalmer could you please review and let us know what do you think?

server/tools/docker-entrypoint.sh
iankko: > why not `set -e` in a separate line?
> Also we probably want `set -eou pipefail`

+1 for changing this into the `set -eou pipefail` form (from Python Zen, when it should fail, fail early and with as much verbosity as possible)

iankko:
> Sure np, just trying to understand. Do you still have issues when docker-entrypoint.sh starts? Do you have any scenario where the script fails but still, the container starts?
> 
> > I can also understand the approach to not fix something that aint broken. But in my opinion this is covered under code quality. I guess if someone used `==` to compare strings in Java and it works for now since the strings are specified in compile time you would probably want to change that to use `equals` instead, right?
> 
> Yes and no, there are several static analysis tools out there. But not everything that's suggested, should be applied to the codebase. Tools and best practices are here to help but in the end, we need to use our best judgment.
> 
> @iankko @douglaspalmer could you please review and let us know what do you think?

AFAICT, all of the aforementioned changes are "just" fixes for the findings reported by ShellCheck.  We used to use that tool regularly (on daily basis, for each PR) in one of the previous projects I was working with (and it had like thousands of shell scripts), and can only recommend to fix those issues. From the experience, those findings sooner or later turned into real bugs.

As for adding the `-e` switch to the shebang / set command, like mentioned above, I would recommend `set -euo pipefail`. This is what we use in Red Hat SSO for OpenShift images too, and can only recommend it. The sooner the program (or image via entrypoint) exhibits the error / failere (within the first testing run already), the better IMHO.

HTH

cartmanume:
> > Sure np, just trying to understand. Do you still have issues when docker-entrypoint.sh starts? Do you have any scenario where the script fails but still, the container starts?
> > > I can also understand the approach to not fix something that aint broken. But in my opinion this is covered under code quality. I guess if someone used `==` to compare strings in Java and it works for now since the strings are specified in compile time you would probably want to change that to use `equals` instead, right?
> > 
> > 
> > Yes and no, there are several static analysis tools out there. But not everything that's suggested, should be applied to the codebase. Tools and best practices are here to help but in the end, we need to use our best judgment.
> > @iankko @douglaspalmer could you please review and let us know what do you think?
> 
> AFAICT, all of the aforementioned changes are "just" fixes for the findings reported by ShellCheck. We used to use that tool regularly (on daily basis, for each PR) in one of the previous projects I was working with (and it had like thousands of shell scripts), and can only recommend to fix those issues. From the experience, those findings sooner or later turned into real bugs.
> 
> As for adding the `-e` switch to the shebang / set command, like mentioned above, I would recommend `set -euo pipefail`. This is what we use in Red Hat SSO for OpenShift images too, and can only recommend it. The sooner the program (or image via entrypoint) exhibits the error / failere (within the first testing run already), the better IMHO.
> 
> HTH

Cool! Thanks for the feedback! I added `set -eou pipefail` now also.

PR #235, opened 2019-11-16
[KEYCLOAK-11696] Follow up to #225: Fix unbound variables

SuperSandro2000:
Fixes
````
Added 'admin' to '/opt/jboss/keycloak/standalone/configuration/keycloak-add-user.json', restart server to load user
/opt/jboss/tools/docker-entrypoint.sh: line 47: KEYCLOAK_FRONTEND_URL: unbound variable
````

PS: I dislike JIRA.

### DIFF
--- a/server/tools/docker-entrypoint.sh
+++ b/server/tools/docker-entrypoint.sh
@@ -9,18 +9,18 @@ file_env() {
     local var="$1"
     local fileVar="${var}_FILE"
     local def="${2:-}"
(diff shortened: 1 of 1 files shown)


stianst:
Awesome thanks for this. Was a bit depressed when first thing I noticed on a Monday morning was the containers where broken, but you made my day with having the fix already ready :)

Found one more problem when testing if DB_VENDOR is not set. Added that to your PR here: https://github.com/keycloak/keycloak-containers/pull/236

PR #237, opened 2019-11-20
KEYCLOAK-12158 Update image in openshift-examples jsons

vramik:


slaskawi:
@vramik @abstractj That's a very good point actually.

To add some context - currently we publish our images to both. However, Quay is preferable.

stianst:
I think it makes sense to change to quay.io/keycloak/keycloak everywhere, but that is a bit bigger task and should probably be discussed on the dev ML. For now I'll merge this though as it fixes the problem.

PR #239, opened 2019-11-29
KEYCLOAK-12306 Do not generate Keystore on empty /etc/x509/https dir

slaskawi:
[KEYCLOAK-12306](https://issues.jboss.org/browse/KEYCLOAK-12306)

The current implementation of the Keycloak Container image generates (and uses!) an empty Keystore if `/etc/x509/https` exists and is empty. This results in making HTTPS port (`8443`) unusable (as there are no common ciphers)


slaskawi:
@stianst @abstractj @iankko Could you plrease review this one when you have a free moment. I would love to get it before `8.0.1` as it hurts Operator + Minikube scenario.

PR #243, opened 2019-12-04
KEYCLOAK-12387 Add jboss group

sventorben:
With KEYCLOAK-10059 the base image has been changed to ubi8-minimal.
Since ubi8-minimal does not ship with a user named 'jboss', the user
will be created in the startup scripts. However, the previously used
image jboss/base-jdk also ships with a group named 'jboss', the
ubi8-minimal image does not. With KEYCLOAK-10059 this group
does not get created

For backwards compatibility a group named 'jboss' has been added and
the user 'jboss' has been associated as a member of that group.

### DIFF
--- a/server/tools/build-keycloak.sh
+++ b/server/tools/build-keycloak.sh
@@ -91,6 +91,7 @@ rm -rf /opt/jboss/keycloak/standalone/configuration/standalone_xml_history
 # Set permissions #
 ###################
 
+echo "jboss:x:1000:jboss" >> /etc/group
 echo "jboss:x:1000:1000:JBoss user:/opt/jboss:/sbin/nologin" >> /etc/passwd
-chown -R jboss:0 /opt/jboss
+chown -R jboss:jboss /opt/jboss
 chmod -R g+rw /opt/jboss


slaskawi:
@stianst @abstractj Looks good to me. Ready to be merged

gt-novelt:
The group has been removed again in 2e22f7bda5f74dcb615012ad8a6360b73b8ae4f5

Any reason for not having the `jboss` group ? I would rather not having the container run with a user having root privileges.
What do you think ?

slaskawi:
@gt-novelt That is correct. We had a lengthy discussion about it in https://github.com/keycloak/keycloak-containers/pull/259 as well in [Keycloak Dev Google Group](https://groups.google.com/forum/#!searchin/keycloak-dev/container%7Csort:date/keycloak-dev/JWUErpYrZAA/Qm1rvjigAwAJ). 

[One of my comments](https://github.com/keycloak/keycloak-containers/pull/259#discussion_r394881223) might give you some hints how not to run the container as `root`.

gt-novelt:
great!

Thank you for all the references.

PR #247, opened 2019-12-13
KEYCLOAK-12947 new docker compose example

ERRECabrera:
added a new example to run keycloak in HA mode with postgres database and jdbc_ping discovery protocol

### DIFF
--- a/docker-compose-examples/README.md
+++ b/docker-compose-examples/README.md
@@ -17,6 +17,19 @@ Open http://localhost:8080/auth and login as user 'admin' with password 'Pa55w0r
 Note - If you run the example twice without removing the persisted volume there will be a warning 'user with username exists'. You can ignore this warning.
 
 
+## Keycloak and PostgreSQL with JDBC_PING
+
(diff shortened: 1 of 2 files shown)


abstractj:
@ERRECabrera thanks for submitting this, although we need a prior discussion about it on the mailing list, plus a Jira. Could you please check https://github.com/keycloak/keycloak-containers/blob/master/CONTRIBUTING.md ?

slaskawi:
@ERRECabrera @abstractj After thinking about this for a while - perhaps we should have just 1 `JDBC_PING` example (@ERRECabrera it might be refactored to Postgresql, I have no preference here). The idea is to show how it works and adding more than 1 example just increases maintenance burden.

slaskawi:
@abstractj @stianst @ERRECabrera What do you think? Do you guys agree with what I've written?

abstractj:
@slaskawi not sure if you started a thread about this, but that would be a fantastic discussion to the dev mailing list.

slaskawi:
Started discussion on this one: https://groups.google.com/forum/#!topic/keycloak-dev/33EiJbMaxlY

PR #254, opened 2020-02-03
[KEYCLOAK-12879] Google identity provider doesn't handle non-ascii characters

hlavki:
This pull request installs `glibc-langpack-en` package because locale files was not installed.

### DIFF
--- a/server/Dockerfile
+++ b/server/Dockerfile
@@ -17,7 +17,7 @@ ARG KEYCLOAK_DIST=https://downloads.jboss.org/keycloak/$KEYCLOAK_VERSION/keycloa
 
 USER root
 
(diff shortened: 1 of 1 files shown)


hlavki:
@slaskawi, image size increase is minimal (+8 MB). I don't think we need another language packs because LC_LANG is set to `en_US.UTF-8`. If I understand it, problem was that LC_LANG was set to locale that wasn't exist.
```
REPOSITORY       TAG     IMAGE ID       CREATED        SIZE
hlavki/keycloak  8.0.1   17dc49695457   17 hours ago   656MB
jboss/keycloak   8.0.1   09676d652352   2 months ago   648MB
```

hlavki:
Anyway I think it would be good to consult solution with creators of redhat universal base image.

slaskawi:
@hlavki This look good to me in general. 

@douglaspalmer @iankko - Perhaps you guys have more expertise in this?

PR #255, opened 2020-02-04
KEYCLOAK-12892 Use dedicated datasource attribute for JDBC_PING

slaskawi:
[KEYCLOAK-12892](https://issues.redhat.com/browse/KEYCLOAK-12892)

Our `JDBC_PING` examples use a generic JNDI Datasource attribute. As mentioned in [WFLY-12948](https://issues.redhat.com/browse/WFLY-12948), we should use a dedicated attribute for this.

### DIFF
--- a/server/tools/cli/jgroups/discovery/JDBC_PING.cli
+++ b/server/tools/cli/jgroups/discovery/JDBC_PING.cli
@@ -1,0 +1,11 @@
+embed-server --server-config=standalone-ha.xml --std-out=echo
+batch
+/subsystem=jgroups/stack=udp/protocol=PING:remove()
+/subsystem=jgroups/stack=udp/protocol=JDBC_PING:add(add-index=0, data-source=KeycloakDS, properties=$keycloak_jgroups_discovery_protocol_properties)
+
+/subsystem=jgroups/stack=tcp/protocol=MPING:remove()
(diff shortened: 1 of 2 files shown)


server/tools/cli/jgroups/discovery/JDBC_PING.cli
```diff
@@ -0,0 +1,11 @@
+embed-server --server-config=standalone-ha.xml --std-out=echo
+batch
+/subsystem=jgroups/stack=udp/protocol=PING:remove()
+/subsystem=jgroups/stack=udp/protocol=JDBC_PING:add(add-index=0, data-source=KeycloakDS, properties=$keycloak_jgroups_discovery_protocol_properties)
```
pedroigor: I remember past discussions where mixing JDBC_PING with the main database could potentially bring issues such as specific permissions (IIRC, Oracle), load, etc. Is this true?

slaskawi: This is a very interesting case.... 

Potentially, you might want to separate those two datasources from each other. They might use different isolation levels (in JGroups only coordinator will write to the DB, whereas in a large Keycloak cluster anyone can write Keycloak database). But in that case, you probably need a handcrafted `standalone.xml` anyway...

I think for containers usecases, where it is super easy to spin up a database, reusing the same datasource might be OK. Unless I don't see any showstoppers...

slaskawi:
@stianst @pedroigor May I ask for a review?

mposolda (approved):
Approving. I did not test this, so I assume you tested it properly that it works as expected with JDBC_PING

PR #256, opened 2020-02-05
[KEYCLOAK-12901] Switch to ubi8-minimal

bruegth:
see https://issues.redhat.com/browse/KEYCLOAK-12901

### DIFF
--- a/gatekeeper/Dockerfile
+++ b/gatekeeper/Dockerfile
@@ -1,4 +1,4 @@
-FROM alpine:3.8
+FROM registry.access.redhat.com/ubi8-minimal
 
 ENV NAME keycloak-gatekeeper
 ENV KEYCLOAK_VERSION 8.0.0
@@ -10,16 +10,20 @@ LABEL Name=keycloak-gatekeeper \
       Url=https://github.com/keycloak/keycloak-gatekeeper \
       Help=https://issues.jboss.org/projects/KEYCLOAK
 
-RUN adduser -D -u 1000 gatekeeper
+USER root
 
-RUN apk add --no-cache ca-certificates curl tar
+WORKDIR "/opt/gatekeeper"
 
-WORKDIR "/opt"
+RUN echo "gatekeeper:x:1000:gatekeeper" >> /etc/group
+RUN echo "gatekeeper:x:1000:1000:gatekeeper user:/opt/gatekeeper:/sbin/nologin" >> /etc/passwd
+RUN chown -R gatekeeper:gatekeeper /opt/gatekeeper && chmod -R g+rw /opt/gatekeeper
 
-RUN curl -fssL "https://downloads.jboss.org/keycloak/$KEYCLOAK_VERSION/gatekeeper/$NAME-$GOOS-$GOARCH.tar.gz" | tar -xz && chmod +x /opt/$NAME
+RUN microdnf update -y && microdnf install -y ca-certificates tar gzip
 
-RUN apk del curl tar
+RUN curl -fssL "https://downloads.jboss.org/keycloak/$KEYCLOAK_VERSION/gatekeeper/$NAME-$GOOS-$GOARCH.tar.gz" | tar -xz && chmod +x /opt/gatekeeper/$NAME
+
+RUN microdnf remove -y tar gzip && microdnf clean all
 
 USER 1000
 
-ENTRYPOINT [ "/opt/keycloak-gatekeeper" ]
+ENTRYPOINT [ "/opt/gatekeeper/keycloak-gatekeeper" ]


gatekeeper/Dockerfile
```diff
@@ -10,15 +10,15 @@ LABEL Name=keycloak-gatekeeper \
       Url=https://github.com/keycloak/keycloak-gatekeeper \
       Help=https://issues.jboss.org/projects/KEYCLOAK
 
-RUN adduser -D -u 1000 gatekeeper
+USER root
```
slaskawi: I think some of our users might depend on existing `gatekeeper` user. @abstractj What do you think about this?

douglaspalmer: In the Keycloak container we use the following to add the jboss user, we should do the same to add the gatekeeper user:

echo "jboss:x:1000:jboss" >> /etc/group
echo "jboss:x:1000:1000:JBoss user:/opt/jboss:/sbin/nologin" >> /etc/passwd

abstractj: @slaskawi I think we should keep the images consistent and stick with what @douglaspalmer suggested.

@bruegth would you mind to update your PR?

bruegth: @abstractj I updated my PR.

I added gatekeeper directory under /opt because make it more consistent to Keycloak and I don't like to make /opt as homedir. Also if someone whats to create a local BoltDB relative to workdir, maybe /opt has not the permissions for that.

gatekeeper/Dockerfile
```diff
@@ -10,16 +10,20 @@ LABEL Name=keycloak-gatekeeper \
...
 
-RUN apk add --no-cache ca-certificates curl tar
+WORKDIR "/opt/gatekeeper"
 
-WORKDIR "/opt"
+RUN echo "gatekeeper:x:1000:gatekeeper" >> /etc/group
```
abstractj: @bruegth Thanks for the changes. I'm not sure about what @slaskawi and @douglaspalmer think about this. But I'd like to avoid, if possible, to have different user from the Keycloak base image. I believe would make things easier for people if we refer to a single user. See:

https://github.com/keycloak/keycloak-containers/blob/9660152ff411ee22cae03c560d0b56aafd875d53/server/tools/build-keycloak.sh#L96

In this way, we keep the references consistent with other images.

Does it make any sense?

iankko: @abstractj @slaskawi @douglaspalmer @bruegth
+1 on keeping the images consistent across the offerings & prefer the `jboss` user similarly like it's done in the base server image.

bruegth: @abstractj @iankko I changed PR to jboss as user. But keep gatekeeper directory below /opt/jboss, please review again.

bruegth: But on other hand: Gatekeeper is based on Go and has nothing to do with `jboss` and the user has only the scope of image where he's created in.

bruegth: Also this will break 
>  One feature/change per PR

 because we change username too.

 So I prefer to revert  back to `gatekeeper `as user, and make new Ticket for the discussion to change user to `jboss`

gatekeeper/Dockerfile
```diff
@@ -10,16 +10,20 @@ LABEL Name=keycloak-gatekeeper \
       Url=https://github.com/keycloak/keycloak-gatekeeper \
       Help=https://issues.jboss.org/projects/KEYCLOAK
 
-RUN adduser -D -u 1000 gatekeeper
+USER root
 
-RUN apk add --no-cache ca-certificates curl tar
+WORKDIR "/opt/jboss/gatekeeper"
```
stianst: Don't use jboss in the WORKDIR.

gatekeeper/Dockerfile
```diff
@@ -10,16 +10,20 @@ LABEL Name=keycloak-gatekeeper \
...
 
-WORKDIR "/opt"
+RUN echo "jboss:x:1000:jboss" >> /etc/group
```
stianst: User shouldn't be jboss, but should be gatekeeper.

gatekeeper/Dockerfile
bruegth: changed

gatekeeper/Dockerfile
bruegth: changed

abstractj:
@stianst feel free to proceed to whatever you think it's best. It's assigned to you now.

gatekeeper/Dockerfile
stianst: The user should not be jboss. This image has nothing to do with jboss. To be honest the keycloak/keycloak image should not use the user jboss either, it should have used the user keycloak.

PR #257, opened 2020-02-07
[KEYCLOAK-12928] Proxy-mappings can be set with an environment variable

vilmosnagy:
This PR creates a new environment variable which will set the [proxy-mappings](https://www.keycloak.org/docs/latest/server_installation/#_proxymappings) to the given value.

### DIFF
--- a/server/tools/change-proxy-mappings.sh
+++ b/server/tools/change-proxy-mappings.sh
@@ -1,0 +1,6 @@
+#!/bin/bash -e
+
+cd /opt/jboss/keycloak
+
+bin/jboss-cli.sh --file=/opt/jboss/tools/cli/proxy-mappings.cli
+rm -rf /opt/jboss/keycloak/standalone/configuration/standalone_xml_history
--- a/server/tools/cli/proxy-mappings.cli
+++ b/server/tools/cli/proxy-mappings.cli
@@ -1,0 +1,6 @@
+embed-server --server-config=standalone.xml --std-out=echo
+echo ** Setting the proxy-mappings **
+/subsystem=keycloak-server/spi=connectionsHttpClient/provider=default:write-attribute(name=properties.proxy-mappings,value=${env.KEYCLOAK_PROXY_MAPPINGS})
+stop-embedded-server
+
+
--- a/server/tools/docker-entrypoint.sh
+++ b/server/tools/docker-entrypoint.sh
@@ -219,6 +219,18 @@ if [ "$DB_VENDOR" != "h2" ]; then
     /bin/sh /opt/jboss/tools/databases/change-database.sh $DB_VENDOR
 fi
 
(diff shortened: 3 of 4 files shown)


server/tools/cli/proxy-mappings.cli
```diff
@@ -0,0 +1,6 @@
+embed-server --server-config=standalone.xml --std-out=echo
+echo ** Setting the proxy-mappings **
+/subsystem=keycloak-server/spi=connectionsHttpClient/provider=default:write-attribute(name=properties.proxy-mappings,value=${env.KEYCLOAK_PROXY_MAPPINGS})
+stop-embedded-server
+
```
slaskawi, line 5: Empty lines that can be removed.

server/tools/docker-entrypoint.sh
```diff
@@ -219,6 +219,18 @@ if [ "$DB_VENDOR" != "h2" ]; then
     /bin/sh /opt/jboss/tools/databases/change-database.sh $DB_VENDOR
 fi
 
+# configure the proxy-mappings
+if [[ -n ${KEYCLOAK_PROXY_MAPPINGS:-} ]]
+then
+    echo "========================================================================="
+    echo ""
+    echo "  Using $KEYCLOAK_PROXY_MAPPINGS proxy-mappings"
```
slaskawi, line 227: I think this could be just a 1 line log message :) Let's see what others think.

server/tools/change-proxy-mappings.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash -e
+
+cd /opt/jboss/keycloak
+
+bin/jboss-cli.sh --file=/opt/jboss/tools/cli/proxy-mappings.cli
+rm -rf /opt/jboss/keycloak/standalone/configuration/standalone_xml_history
```
slaskawi, line 6: This line should be removed. It doesn't belong here.

stianst:
I would probably only add this if there are others that want it as well. We can't really expose all config as environment variables as its not really maintainable, at some point you'll need to do something custom to configure.

In Keycloak.X this issue will be resolved as all config is effectively doable through environment variables out of the box.

slaskawi:
I created a discussion topic for this: https://groups.google.com/forum/#!topic/keycloak-dev/a3fzQUuOw48

slaskawi:
I agree with @stianst in [Keycloak DEV](https://groups.google.com/d/msg/keycloak-dev/a3fzQUuOw48/0-PtxAD5AgAJ):

> I think we should reject this PR. As I mentioned it simply doesn't scale to add all config options as environment variables to the existing image, and in the future this problem will be solved in Keycloak.X as all config options are automatically exposed as environment variables.

Let's close this PR for now and get back to it once Keycloak.X is released.

PR #258, opened 2020-02-24
KEYCLOAK-13020 Permissions update

slaskawi:
[KEYCLOAK-13020](https://issues.redhat.com/browse/KEYCLOAK-13020)

This Pull Request sets up permissions to all files in `/opt/jboss`. You may play with it using this image: `docker.io/slaskawi/keycloak:KEYCLOAK-13020`

### DIFF
--- a/server/tools/build-keycloak.sh
+++ b/server/tools/build-keycloak.sh
@@ -87,6 +87,13 @@ rm -rf /opt/jboss/keycloak/standalone/configuration/standalone_xml_history
 bin/jboss-cli.sh --file=/opt/jboss/tools/cli/standalone-ha-configuration.cli
 rm -rf /opt/jboss/keycloak/standalone/configuration/standalone_xml_history
 
+###########
+# Garbage #
+###########
+
+rm -rf /opt/jboss/keycloak/standalone/tmp/auth
+rm -rf /opt/jboss/keycloak/domain/tmp/auth
+
 ###################
 # Set permissions #
 ###################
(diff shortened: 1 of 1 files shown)


server/tools/build-keycloak.sh
```diff
@@ -87,14 +87,29 @@ rm -rf /opt/jboss/keycloak/standalone/configuration/standalone_xml_history
...
 echo "jboss:x:1000:jboss" >> /etc/group
 echo "jboss:x:1000:1000:JBoss user:/opt/jboss:/sbin/nologin" >> /etc/passwd
 chown -R jboss:jboss /opt/jboss
-chmod -R g+rw /opt/jboss
-
-rm -rf /opt/jboss/keycloak/standalone/tmp/auth
-rm -rf /opt/jboss/keycloak/domain/tmp/auth
+chmod -R 744 /opt/jboss
+# Enable listings
+chmod -R +X /opt/jboss
```
iankko: @slaskawi Is there a reason, the `X` modifier in the command above needs to be in capital letter? Confirmed it works too, but wondering about the motivation, why it was chosen?

slaskawi: The `X` parameter applies executable permission only to directories. This is exactly what we want to achieve here.

slaskawi:
> Is the lack of trailing newline intentional, or just an oversight? I'm sure it's harmless either way, just curious.

@leehambley This was my mistake :) Thanks for catching this!

> I've always felt that our permission were already too permissive. Why does anyone other than the > jboss user need to be able to write or execute anything? Would it not suffice to replace:
> 
> chmod -R g+rw /opt/jboss
> 
> with:
> 
> chmod -R g+rw,o+r /opt/jboss

@douglaspalmer That's an interesting hypothesis. Here's a nice write up from [OpenShift 3.3 manual](https://docs.openshift.com/container-platform/3.3/creating_images/guidelines.html):

> For an image to support running as an arbitrary user, directories and files that may be written to by processes in the image should be owned by the root group and be read/writable by that group. Files to be executed should also have group execute permissions.
> Adding the following to your Dockerfile sets the directory and file permissions to allow users in the root group to access them in the built image:
> RUN chgrp -R 0 /some/directory && \
>    chmod -R g=u /some/directory
...
> Because the container user is always a member of the root group, the container user can read and write these files. The root group does not have any special permissions (unlike the root user) so there are no security concerns with this arrangement. In addition, the processes running in the container must not listen on privileged ports (ports below 1024), since they are not running as a privileged user.

So let me check if that works and get back to you shortly.

slaskawi:
@douglaspalmer I ran some experiments and it seems it is possible to implement this in a more restricted way. See https://github.com/keycloak/keycloak-containers/pull/259

EagleIJoe:
This works well for me.

slaskawi:
This Pull Request solves [KEYCLOAK-13333](https://issues.redhat.com/browse/KEYCLOAK-13333)

slaskawi:
I'm closing this in favor of https://github.com/keycloak/keycloak-containers/pull/259 (the more restrictive model). If we hit any problems, we'll fallback to this one.

PR #259, opened 2020-02-25
KEYCLOAK-13020 Permissions update

slaskawi:
[KEYCLOAK-13020](https://issues.redhat.com/browse/KEYCLOAK-13020)

This Pull Request uses a more restrictive way of setting permissions proposed by @douglaspalmer in https://github.com/keycloak/keycloak-containers/pull/258. 

You may play with it using this image: `docker.io/slaskawi/keycloak:KEYCLOAK-13020-2`

Unfortunately there are some cons to this approach. It doesn't fully fix [KEYCLOAK-13020](https://issues.redhat.com/browse/KEYCLOAK-13020). Users will always need to run this container with a user, who uses group id equal to `0` (root). This way, they will simulate, how OpenShift does it: https://docs.openshift.com/container-platform/3.3/creating_images/guidelines.html

### DIFF
--- a/server/tools/build-keycloak.sh
+++ b/server/tools/build-keycloak.sh
@@ -87,14 +87,18 @@ rm -rf /opt/jboss/keycloak/standalone/configuration/standalone_xml_history
 bin/jboss-cli.sh --file=/opt/jboss/tools/cli/standalone-ha-configuration.cli
 rm -rf /opt/jboss/keycloak/standalone/configuration/standalone_xml_history
 
+###########
+# Garbage #
+###########
+
+rm -rf /opt/jboss/keycloak/standalone/tmp/auth
+rm -rf /opt/jboss/keycloak/domain/tmp/auth
+
 ###################
 # Set permissions #
 ###################
 
-echo "jboss:x:1000:jboss" >> /etc/group
-echo "jboss:x:1000:1000:JBoss user:/opt/jboss:/sbin/nologin" >> /etc/passwd
-chown -R jboss:jboss /opt/jboss
-chmod -R g+rw /opt/jboss
-
-rm -rf /opt/jboss/keycloak/standalone/tmp/auth
-rm -rf /opt/jboss/keycloak/domain/tmp/auth
+echo "jboss:x:0:root" >> /etc/group
+echo "jboss:x:1000:0:JBoss user:/opt/jboss:/sbin/nologin" >> /etc/passwd
+chown -R jboss:root /opt/jboss
+chmod -R g+rwX /opt/jboss


leehambley:
I'm checking now, but this may not allow people using Docker with mapped user namespaces:

- https://docs.docker.com/engine/security/userns-remap/

leehambley:
This works (boots, and I can get to the login, I didn't test further) with mapped userns,

server/tools/build-keycloak.sh
```diff
@@ -87,14 +87,20 @@ rm -rf /opt/jboss/keycloak/standalone/configuration/standalone_xml_history
...
-echo "jboss:x:1000:1000:JBoss user:/opt/jboss:/sbin/nologin" >> /etc/passwd
-chown -R jboss:jboss /opt/jboss
-chmod -R g+rw /opt/jboss
-
-rm -rf /opt/jboss/keycloak/standalone/tmp/auth
-rm -rf /opt/jboss/keycloak/domain/tmp/auth
+echo "jboss:x:0:root" >> /etc/group
+echo "jboss:x:1000:0:JBoss user:/opt/jboss:/sbin/nologin" >> /etc/passwd
+chown -R jboss:root /opt/jboss
+chmod -R 774 /opt/jboss
```
douglaspalmer: This sets execute permissions on text files for owner and group. Why is this better than 'chmod -R g+rw,o+r /opt/jboss'?

douglaspalmer: Actually, it would be better to just use 'chmod -R g=u /opt/jboss' so that the group has the same permissions as the user.

douglaspalmer: You could use 'chmod g=u,o+r /opt/jboss' if you really think other needs read permissions. I would prefer 'chmod g=u,o= /opt/jboss'. If we agree that other doesn't need access then you can also remove the 'chmod -R +X /opt/jboss'.

slaskawi: I've just tried your suggested approach:
```
echo "jboss:x:0:root" >> /etc/group
echo "jboss:x:1000:0:JBoss user:/opt/jboss:/sbin/nologin" >> /etc/passwd
chown -R jboss:root /opt/jboss
chmod g=u,o= /opt/jboss
```

Unfortunately the container crashes (the container is pushed to dockerhub so you can try it yourself):

```
$ docker run --user 123:0 docker.io/slaskawi/keycloak:KEYCLOAK-13020-3
(...)
Unable to read the logging configuration from 'file:/opt/jboss/keycloak/standalone/configuration/logging.properties' (java.io.FileNotFoundException: /opt/jboss/keycloak/standalone/configuration/logging.properties (Permission denied))
java.util.concurrent.ExecutionException: JBTHR00005: Operation failed
	at org.jboss.threads@2.3.3.Final//org.jboss.threads.AsyncFutureTask.get(AsyncFutureTask.java:253)
	at org.jboss.as.server@10.0.0.Final//org.jboss.as.server.Main.main(Main.java:109)
	at org.jboss.modules.Module.run(Module.java:352)
	at org.jboss.modules.Module.run(Module.java:320)
	at org.jboss.modules.Main.main(Main.java:593)
Caused by: org.jboss.msc.service.StartException in service jboss.as: Failed to start service
```

In general, we can not assume that we will always be running the container as `jboss` (or `root`) user. OpenShift gives you a random user belonging to `root` group. I think, we should go one step further and allow running this container as a random user belonging to a random group. In other words, I'm slightly leaning towards https://github.com/keycloak/keycloak-containers/pull/258

slaskawi:
This Pull Request solves [KEYCLOAK-13333](https://issues.redhat.com/browse/KEYCLOAK-13333)

leehambley:
I just saw #258 was closed in favour of this. :+1: 

For clarity, this image (docker.io/slaskawi/keycloak:KEYCLOAK-13020-2) works perfectly under Docker with a mapped userns, the original (`KEYCLOAK-13020`) issue I reported.

slaskawi:
Thank you for checking this, @leehambley !

server/tools/build-keycloak.sh
iankko: @douglaspalmer @slaskawi Though per [this comment](https://github.com/openshift/origin/issues/18974#issuecomment-374578684) it's possible to run the pod under the same UID in OpenShift too, not convinced we should follow the use of that.

But what about using the [recommended form](https://docs.openshift.com/container-platform/3.11/creating_images/guidelines.html#openshift-specific-guidelines), see `Support Arbitrary User IDs` in that section.

Would the
```
RUN chgrp -R 0 /some/directory && \
    chmod -R g+rwX /some/directory
```

form, suggested there, help?

iankko: @slaskawi Confirmed the change works fine with both `OCP v4.3.5` & `OCP vv3.11.154`. But like mentioned earlier, I would suggest to change the `/opt/jboss` group ownership & mode to follow the official guide. IOW cherrypick [this commit](https://github.com/iankko/keycloak-1/commit/b717eb7d8f14faaa662436e2a5de2471206ecd08). That one works on (both versions too) & follows the guideline.

slaskawi: I added suggested changes and tested them using this Docker image: `docker.io/slaskawi/keycloak:KEYCLOAK-13020-3`. Here is a short list of use cases I tested:

* [x] Installed Docker on Ubuntu Docker image and tested mapped user namespaces as suggested in [KEYCLOAK-13310](https://issues.redhat.com/browse/KEYCLOAK-13310)
* [x] Tested running as an arbitrary user `docker run --user 123 docker.io/slaskawi/keycloak:KEYCLOAK-13020-3`
* [x] Tested running as an arbitrary user belonging to an non-root group (this fails, but it's expected): `docker run --user 123:3 docker.io/slaskawi/keycloak:KEYCLOAK-13020-3`
* [ ] Testing on OpenShift 4 at the moment.

server/tools/build-keycloak.sh
```diff
@@ -87,14 +87,19 @@ rm -rf /opt/jboss/keycloak/standalone/configuration/standalone_xml_history
...
 ###################
 
-echo "jboss:x:1000:jboss" >> /etc/group
-echo "jboss:x:1000:1000:JBoss user:/opt/jboss:/sbin/nologin" >> /etc/passwd
-chown -R jboss:jboss /opt/jboss
-chmod -R g+rw /opt/jboss
-
-rm -rf /opt/jboss/keycloak/standalone/tmp/auth
-rm -rf /opt/jboss/keycloak/domain/tmp/auth
+echo "jboss:x:0:root" >> /etc/group
```
douglaspalmer: If the jboss user is going to be in the root group we no longer need to create a jboss group and the group id 0 is already assigned to the root group.

server/tools/build-keycloak.sh
```diff
@@ -87,14 +87,19 @@ rm -rf /opt/jboss/keycloak/standalone/configuration/standalone_xml_history
...
+echo "jboss:x:1000:0:JBoss user:/opt/jboss:/sbin/nologin" >> /etc/passwd
+chown -R jboss:root /opt/jboss
+chgrp -R 0 /opt/jboss
```
douglaspalmer: This chgrp is unnecessary as the chown on the previous line has already changed the group to root (0).

server/tools/build-keycloak.sh
douglaspalmer: This would add the jboss user to the root group: 
sed -e '1s/$/jboss/' /etc/group

slaskawi:
Hey @stianst @douglaspalmer @iankko ,

Having in mind our very limited bandwidth, and the code freeze, that happens to be today, I vote to merge this one as is and create a follow up JIRA to implement changes suggested by @douglaspalmer . Does it sound OK with you?

Thanks,
Sebastian

iankko:
> Hey @stianst @douglaspalmer @iankko ,
> 
> Having in mind our very limited bandwidth, and the code freeze, that happens to be today, I vote to merge this one as is and create a follow up JIRA to implement changes suggested by @douglaspalmer . Does it sound OK with you?

@slaskawi 

Not ideal, but what else can be done @ this time (getting the latest version of Keycloak containers back working on OCP 4.x seems pretty urgent need)

So +1 for following the procedure you suggested above for now from my PoV

> 
> Thanks,
> Sebastian

HTH
Jan

server/tools/build-keycloak.sh
slaskawi: Fixed.

stianst:
@douglaspalmer @slaskawi merging this as it looks fine to me, and we need this included in 9.0.1. If we need to improve anything please create a follow-up task for that.

slaskawi:
@douglaspalmer @iankko @sventorben - when working on this issue with Stian, we started to question the need for "jboss" user. I created a Keycloak Dev discussion to get an agreement whether we want to keep it or remove it. Please have a look at it here: https://groups.google.com/forum/#!topic/keycloak-dev/JWUErpYrZAA

srueg:
This bug just hit me on OpenShift (permission denied). Is there an ETA for a new container image build including this fix?

server/tools/build-keycloak.sh
douglaspalmer: This line creates a group named 'jboss', with the same group id as the 'root' group, with one group member and that member is 'root'. The jboss user isn't a member of the root group.

douglaspalmer: Since the jboss user is assigned the root group from /etc/passwd we should probably just delete this line altogether.

douglaspalmer: After a bit of digging it seems like the jboss group is simply ignored. Executing 'chgrp jboss aFile' will cause aFile to be in the 'root' group.

PR #262, opened 2020-03-19
Add properties to enable upload script during import

tbocinec:
Importing from an app does not work without uploading scripts.


slaskawi:
@pedroigor May I ask you to look into this?

stianst:
Enabling a feature that was deprecated due to security concerns to resolve an bug in import is not the correct way to solve anything. I also believe this issue has been resolved.

PR #264, opened 2020-04-01
WIP: Fix keycloak container restart issues

iavael:
When container restarts these issues occur:
* scripts (x509.sh, jgroups.sh etc) append configuration to .jbossclirc again
* jdbc-driver is attempted to be added even if it's already present in standalone-ha.xml
* x509.sh tries to access existsing truststores with new random passwords and append certs/keys on each container start

### DIFF
--- a/server/tools/cli/databases/mariadb/change-database.cli
+++ b/server/tools/cli/databases/mariadb/change-database.cli
@@ -6,4 +6,6 @@
 /subsystem=datasources/data-source=KeycloakDS: write-attribute(name=background-validation, value=true)
 /subsystem=datasources/data-source=KeycloakDS: write-attribute(name=background-validation-millis, value=60000)
 /subsystem=datasources/data-source=KeycloakDS: write-attribute(name=flush-strategy, value=IdleConnections)
-/subsystem=datasources/jdbc-driver=mariadb:add(driver-name=mariadb, driver-module-name=org.mariadb.jdbc, driver-xa-datasource-class-name=org.mariadb.jdbc.MySQLDataSource)
+
+/subsystem=datasources/jdbc-driver=mariadb: remove()
+/subsystem=datasources/jdbc-driver=mariadb: add(driver-name=mariadb, driver-module-name=org.mariadb.jdbc, driver-xa-datasource-class-name=org.mariadb.jdbc.MySQLDataSource)
--- a/server/tools/docker-entrypoint.sh
+++ b/server/tools/docker-entrypoint.sh
@@ -219,6 +219,8 @@ if [ "$DB_VENDOR" != "h2" ]; then
     /bin/sh /opt/jboss/tools/databases/change-database.sh $DB_VENDOR
 fi
 
+sed -i '/^set keycloak/d' "${JBOSS_HOME}/bin/.jbossclirc"
+
 /opt/jboss/tools/x509.sh
 /opt/jboss/tools/jgroups.sh
 /opt/jboss/tools/infinispan.sh
--- a/server/tools/x509.sh
+++ b/server/tools/x509.sh
@@ -25,6 +25,8 @@ function autogenerate_keystores() {
 
       echo "Creating ${KEYSTORES[$KEYSTORE_TYPE]} keystore via OpenShift's service serving x509 certificate secrets.."
 
+      rm -f "${JKS_KEYSTORE_FILE}" "${PKCS12_KEYSTORE_FILE}"
+
       openssl pkcs12 -export \
       -name "${NAME}" \
       -inkey "${X509_KEYSTORE_DIR}/${X509_KEY}" \
@@ -64,6 +66,9 @@ function autogenerate_keystores() {
   if [ -n "${X509_CA_BUNDLE}" ]; then
     pushd /tmp >& /dev/null
     echo "Creating Keycloak truststore.."
+
+    rm -f "${JKS_TRUSTSTORE_PATH}"
+
     # We use cat here, so that users could specify multiple CA Bundles using space or even wildcard:
     # X509_CA_BUNDLE=/var/run/secrets/kubernetes.io/serviceaccount/*.crt
     # Note, that there is no quotes here, that's intentional. Once can use spaces in the $X509_CA_BUNDLE like this:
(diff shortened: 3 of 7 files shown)


server/tools/cli/databases/mariadb/change-database.cli
```diff
@@ -6,4 +6,6 @@
 /subsystem=datasources/data-source=KeycloakDS: write-attribute(name=background-validation, value=true)
 /subsystem=datasources/data-source=KeycloakDS: write-attribute(name=background-validation-millis, value=60000)
 /subsystem=datasources/data-source=KeycloakDS: write-attribute(name=flush-strategy, value=IdleConnections)
-/subsystem=datasources/jdbc-driver=mariadb:add(driver-name=mariadb, driver-module-name=org.mariadb.jdbc, driver-xa-datasource-class-name=org.mariadb.jdbc.MySQLDataSource)
+
+/subsystem=datasources/jdbc-driver=mariadb: remove()
```
stianst, line 10: jboss-cli supports conditions, so instead of remove/add you can just check if it's already there.

server/tools/docker-entrypoint.sh
```diff
@@ -219,6 +219,8 @@ if [ "$DB_VENDOR" != "h2" ]; then
     /bin/sh /opt/jboss/tools/databases/change-database.sh $DB_VENDOR
 fi
 
+sed -i '/^set keycloak/d' "${JBOSS_HOME}/bin/.jbossclirc"
```
stianst, line 222: Got no clue what this is doing

server/tools/x509.sh
```diff
@@ -25,6 +25,8 @@ function autogenerate_keystores() {
 
       echo "Creating ${KEYSTORES[$KEYSTORE_TYPE]} keystore via OpenShift's service serving x509 certificate secrets.."
 
+      rm -f "${JKS_KEYSTORE_FILE}" "${PKCS12_KEYSTORE_FILE}"
```
stianst, line 28: Should really just check if this was already done instead of re-doing it

stianst:
Please create a JIRA with a proper description, and also please remember we do require commits to be squashed to a single commit.

server/tools/cli/databases/mariadb/change-database.cli
iavael: Yeah, I read about this feature after I commited this. Wanted to use "if" here, but it doesn't work. Maybe you can help me?
When I write
```
if (outcome == success) of /subsystem=datasources/jdbc-driver=postgresql:read-resource()
  /subsystem=datasources/jdbc-driver=postgresql: remove()
end-if
```
I see this in keycloak log
```
The batch failed with the following error: : Failed to create batch from /opt/jboss/tools/cli/databases/postgres/change-database.cli: java.util.concurrent.ExecutionException: org.jboss.as.cli.CommandFormatException: The command is not available in the current context (e.g. required subsystems or connection to the controller might be unavailable).: org.jboss.as.cli.CommandFormatException: The command is not available in the current context (e.g. required subsystems or connection to the controller might be unavailable).: The command is not available in the current context (e.g. required subsystems or connection to the controller might be unavailable).
```

server/tools/docker-entrypoint.sh
iavael: This command remove all strings starting with "set keycloak". This is done, because they gonna be unconditionally appended again after that.

server/tools/x509.sh
iavael: It's not trivial, because existing keystores were created with random passwords. These passwords are saved in standalone-ha.xml, but reading them from it looks too complex. And handling changes of certs and keys would add even more complexity.

iavael:
@stianst I do remember about squashing commits and I was going to create jira task after I finnaly figure out why keycloak container crashes after restart. Changes in this PR fix some issues, but problem is still exists, so this PR is WIP

server/tools/cli/databases/mariadb/change-database.cli
slaskawi: I totally agree with @stianst here. I think the whole script should be conditional (not just adding database driver).

server/tools/x509.sh
slaskawi: Running this whole script should be conditional as well. If it has been ran, do not do it again.

iavael: @slaskawi 
How certificate is supposed to be updated then?

server/tools/cli/databases/mariadb/change-database.cli
iavael: I agree about running scripts, that rely only on configuration from environment variables, which cannot be changed for existing container.
But I don't think that doing this for scripts based on mutable configuration from volumes (like x509-related script or custom cli-scripts) is a valid approach.

kplantjr:
I have a pretty hacky solution that seems to work for me, making the following changes in https://github.com/keycloak/keycloak-containers/blob/10.0.0/server/tools/docker-entrypoint.sh#L218-L227

```
ALREADY_STARTED="/opt/jboss/ALREADY_STARTED"
if [ ! -e $ALREADY_STARTED ]; then
        touch $ALREADY_STARTED
        ALREADY_STARTED=0
        if [ "$DB_VENDOR" != "h2" ]; then
            /bin/sh /opt/jboss/tools/databases/change-database.sh $DB_VENDOR
        fi
        /opt/jboss/tools/x509.sh
        /opt/jboss/tools/jgroups.sh
fi

/opt/jboss/tools/infinispan.sh
/opt/jboss/tools/statistics.sh
/opt/jboss/tools/autorun.sh
/opt/jboss/tools/vault.sh
```

slaskawi:
@kplantjr Could you please tell me what is the status of this Pull Request?

I believe we have two options in moving this forward:
- Make all CLI scripts conditional (e.g. check if Maria DB Driver is already there before adding it). This option would be my preference.
- Make all shell scripts conditional, as you proposed [here](https://github.com/keycloak/keycloak-containers/pull/264#issuecomment-626659464). Even though I don't like this approach, it's definitely simpler to start with.

iavael:
@slaskawi second approach is not that simple, if you want user to be able to update ssl-cert without deleting container.

slaskawi:
@iavael Modifying SSL key/certs is something that should be out of the scope. Container should be immutable by their nature. If you'd like the new certs to be picked up, you should always restart the container (by restart I mean kill it and run again, not pause/resume).

iavael:
@slaskawi that's a bit counterintuitive considering that ssl cert/key are provided via host volumes. As a user I'd expect, that if I update ssl cert on the host and restart container with respectful directory "volumed", then new ssl cert should be caught up by keycloak, because this state is actually out of container scope.

slaskawi:
I started a discussion about this here: https://groups.google.com/forum/#!topic/keycloak-dev/a3fzQUuOw48

slaskawi:
As a result of [Keycloak DEV discussion](https://groups.google.com/d/msg/keycloak-dev/I3qfqTtVmvw/j5RtiBb5AgAJ), I'm closing this one.

An explicit statement about `docker start` and `stop` will be added in https://github.com/keycloak/keycloak-containers/pull/276

PR #266, opened 2020-04-24
update tls config readme to clarify usage with container tools

JGjorgji:
I see that a lot of people were having this issue but with no resolution, what happens is by default the key generated by lets encrypt tools is writen as `700` which is correct. However when mounting the cert/key into the container, the container tools mount the files as root, meaning when the container is run as the jboss user it the `x509.sh` script cannot read the key and generates an empty keystore.
This makes the container start up fine but they you get weird errors like this:
```
error:14094410:SSL routines:ssl3_read_bytes:sslv3 alert handshake failure
```


slaskawi:
@stianst This one LGTM!

JGjorgji:
> @stianst This one LGTM!

Just to let you know i don't have write access to this so you or someone else will have to merge this.

PR #269, opened 2020-05-04
KEYCLOAK-12030 Convert init-container script from Python to Bash

hmlnarik:


hmlnarik:
@pb82 Could you please check that this covers your use cases just the same as the original Python script?

slaskawi:
@hmlnarik The code LGTM, however I would like to test it out in the Operator. Would you mind to push the image to some public registry?

pb82:
@hmlnarik It looks like everything the python code did is covered, although i haven't tried it with keycloak yet.

hmlnarik:
In addition I have removed platform python as this is not reasonable for init container and inflates the image size by one third (not to speak for additional potential for respin induced by packages that are removed).
The built image is available via `podman pull quay.io/hmlnarik/keycloak-init-container:269.1.0`

PR #270, opened 2020-05-12
KEYCLOAK-12030 Productization script

slaskawi:
This Pull Request introduces productization script for Keycloak/RHSSO Init Container

### DIFF
--- a/keycloak-init-container/productize.sh
+++ b/keycloak-init-container/productize.sh
@@ -1,0 +1,20 @@
+#!/bin/bash -ex
+
+LABELS=$(cat <<-END
+LABEL \
+com.redhat.component="redhat-sso-7-sso74-init-container-rhel8-container"  \
+description="Red Hat Single Sign-On 7.4 Init container image" \
+summary="Red Hat Single Sign-On 7.4 Init container image" \
+version="7.4" \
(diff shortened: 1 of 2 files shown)


keycloak-init-container/productize.sh
```diff
@@ -0,0 +1,20 @@
...
+io.k8s.description="Init Container for Red Hat SSO" \
+io.k8s.display-name="Red Hat SSO 7.4 Init Container" \
+io.openshift.tags="sso,sso74,keycloak,operator" \
+name="rh-sso-7\/sso74-init-container-rhel8-container" \
+maintainer="Red Hat Single Sign-On Team"
+END
+)
+
+sed -i 's/registry.access.redhat.com\/ubi8-minimal/ubi8-minimal:8-released/' Dockerfile
+sed -i -e 's/FROM registry.access.redhat.com\/ubi8\/ubi-minimal:[0-9.]*/FROM ubi8-minimal:8-released/' \
```
ASzc, line 18: FROM for init uses `registry.access.redhat.com/ubi8/python-36` Not sure what the equiv. internally is.

slaskawi: That's wrong I think, isn't that @hmlnarik ? We removed the Python script, so now we should be layered on UBI.

PR #277, opened 2020-07-11
KEYCLOAK-14712 Support for connecting to MSSQL using NTLM

antaran13:
This pull request should enable connecting to MSSQL with NTLM by

1. updating JDBC driver to the newest stable version 8.2.2
2. not appending question mark to JDBC_PARAMS when database is MSSQL
3. not hardcoding integratedSecurity=false in CLI script (false is default value anyway)

Jira: https://issues.redhat.com/browse/KEYCLOAK-14712
Mailing list: https://groups.google.com/forum/#!topic/keycloak-dev/XoVdG2r_e3o


server/tools/cli/databases/mssql/change-database.cli
```diff
@@ -1,5 +1,5 @@
 /subsystem=datasources/data-source=KeycloakDS: remove()
-/subsystem=datasources/data-source=KeycloakDS: add(jndi-name=java:jboss/datasources/KeycloakDS,enabled=true,use-java-context=true,use-ccm=true, connection-url="jdbc:sqlserver://${env.DB_ADDR:mssql}:${env.DB_PORT:1433};databaseName=${env.DB_DATABASE:keycloak};sendStringParametersAsUnicode=false;integratedSecurity=false;user=${env.DB_USER:keycloak};password=${env.DB_PASSWORD:password};${env.JDBC_PARAMS:}", driver-name=sqlserver)
+/subsystem=datasources/data-source=KeycloakDS: add(jndi-name=java:jboss/datasources/KeycloakDS,enabled=true,use-java-context=true,use-ccm=true, connection-url="jdbc:sqlserver://${env.DB_ADDR:mssql}:${env.DB_PORT:1433};databaseName=${env.DB_DATABASE:keycloak};sendStringParametersAsUnicode=false;user=${env.DB_USER:keycloak};password=${env.DB_PASSWORD:password};${env.JDBC_PARAMS:}", driver-name=sqlserver)
```
sschu, line 2: I am wondering if this is a compatible change. What happens if "integratedSecurity=false" is not explicitely specified. Is it the default? If not, this change might be breaking for existing users.

ibauersachs: I have never specified `integratedSecurity=false` in an MSSQL connection string. And from looking at the [source](https://github.com/microsoft/mssql-jdbc/blob/v8.2.2/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerDriver.java#L408), `false` is indeed the default.

antaran13: Possible properties with explanation and default values are also listed here: [setting-the-connection-properties](https://docs.microsoft.com/en-us/sql/connect/jdbc/setting-the-connection-properties?view=sql-server-ver15)

Just search for `integratedSecurity`.

I have just realized that this change should not only enable to set up NTLM, but Kerberos as well.

sschu: Sounds good, thanks.

stianst (approved):

stianst:
@abstractj LGTM - since it's been two days since you sent out the mail on dev ml, and this one has been around for some time, I'd say it's ready to be merged

sschu:
Maybe the driver version could be pushed to 8.4.0 as thats currently the latest stable? https://docs.microsoft.com/en-us/sql/connect/jdbc/download-microsoft-jdbc-driver-for-sql-server?view=sql-server-ver15

antaran13:
> Maybe the driver version could be pushed to 8.4.0 as thats currently the latest stable? https://docs.microsoft.com/en-us/sql/connect/jdbc/download-microsoft-jdbc-driver-for-sql-server?view=sql-server-ver15

At the time when this PR was created latest stable version was 8.2.2 and that is what I am using right now and what was tested.

Still there is one remaining issue I would like to fix after this PR is closed so I can test the newest driver while working on it.
Therefore I would rather create separate PR for the newest version after this one is merged.

PR #281, opened 2020-07-30
x509.sh: add message about key permissions

thomasdstewart:
Add message about key permissions if ${KEYSTORES_STORAGE}/${JKS_KEYSTORE_FILE} is not created.

When starting with something like "docker run .... -v .../key.pem:/etc/x509/https/tls.key quay.io/keycloak/keycloak:latest" if the key (${KEYSTORES_STORAGE}/${JKS_KEYSTORE_FILE}) is not created, for some reasons including permissions issues, no message is displayed and SSL is started in a broken state. This shows a warning message to help identify the issue.

### DIFF
--- a/server/tools/x509.sh
+++ b/server/tools/x509.sh
@@ -41,6 +41,8 @@ function autogenerate_keystores() {
 
       if [ -f "${KEYSTORES_STORAGE}/${JKS_KEYSTORE_FILE}" ]; then
         echo "${KEYSTORES[$KEYSTORE_TYPE]} keystore successfully created at: ${KEYSTORES_STORAGE}/${JKS_KEYSTORE_FILE}"
+      else
+        echo "${KEYSTORES[$KEYSTORE_TYPE]} keystore not created at: ${KEYSTORES_STORAGE}/${JKS_KEYSTORE_FILE} (check permissions?)"
       fi
 
       echo "set keycloak_tls_keystore_password=${PASSWORD}" >> "$JBOSS_HOME/bin/.jbossclirc"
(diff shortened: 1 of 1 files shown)


server/tools/x509.sh
```diff
@@ -41,6 +41,8 @@ function autogenerate_keystores() {
 
       if [ -f "${KEYSTORES_STORAGE}/${JKS_KEYSTORE_FILE}" ]; then
         echo "${KEYSTORES[$KEYSTORE_TYPE]} keystore successfully created at: ${KEYSTORES_STORAGE}/${JKS_KEYSTORE_FILE}"
+      else
+        echo "${KEYSTORES[$KEYSTORE_TYPE]} keystore not created at: ${KEYSTORES_STORAGE}/${JKS_KEYSTORE_FILE} (check permissions?)"
```
slaskawi, line 45: Perhaps we should remove the `(check permissions?)` part. There might be some other reasons why this file wasn't create (e.g. not enough space on the disk).

Also, could you please add a similar change a bit lower - in the trust store part?

thomasdstewart:
Hi @slaskawi, I agree that there could be other reasons why the file is not created. However I would argue that because it's dealing with a ssl key that file read permissions is a common issue and the only way round this issue is to dig around in the guts of Dockerfiles to find out what magic is going on under the covers. The idea is to create less friction when new users try out keycloak via docker, eg say someone follows https://www.keycloak.org/getting-started/getting-started-docker, the very next step which is also suggested is "Configure SSL with your own certificates", which is indeed what I went on to do, in order to test a less trivial application. However when googeling for "site:www.keycloak.org inurl:docs ssl" it leads to https://www.keycloak.org/docs/latest/server_installation/#_setting_up_ssl which does not document the docker options (yet). Others have run into this with little success over the root cause, eg https://stackoverflow.com/questions/59438777/keycloak-error-with-docker-and-letsencrypt-certificates so I would suggest leaving the albeit ugly check permissions message there. Also I've added a second similar message for the trust store in the latest commit. :)

slaskawi:
Works for me @thomasdstewart !

slaskawi:
@thomasdstewart The code LGTM. Can I ask you to squash your commits and add a JIRA for this? Once you do this, I'll approve and merge this PR.

thomasdstewart:
Done :)

slaskawi:
@thomasdstewart Integrated. Thanks!

PR #282, opened 2020-07-31
KEYCLOAK-14495 Introduce support for JAVA_OPTS_APPEND

radudd:
This is a proposal to introduce support for JAVA_OPTS_APPEND environment variable. The corresponding JIRA ticket: https://issues.redhat.com/browse/KEYCLOAK-14495

### DIFF
--- a/server/tools/docker-entrypoint.sh
+++ b/server/tools/docker-entrypoint.sh
@@ -104,6 +104,9 @@ if echo "$@" | grep -E -v -- '-c |-c=|--server-config |--server-config='; then
     SYS_PROPS+=" -c=standalone-ha.xml"
 fi
 
+# Adding support for JAVA_OPTS_APPEND
+sed -i '$a\\n# Append to JAVA_OPTS. Necessary to prevent some values being omitted if JAVA_OPTS is defined directly\nJAVA_OPTS=\"\$JAVA_OPTS \$JAVA_OPTS_APPEND\"' /opt/jboss/keycloak/bin/standalone.conf
+
 ############
 # DB setup #
 ############
--- a/server/README.md
+++ b/server/README.md
@@ -306,6 +306,17 @@ Or you can volume the entire directory to supply a directory of scripts.
 Note that when combining the approach of extending the image and `volume`ing the entire directory, the volume will override
 all scripts shipped in the image.
 
+## Start a Keycloak instance with custom command-line options
+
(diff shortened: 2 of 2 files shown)


server/tools/docker-entrypoint.sh
```diff
@@ -104,6 +104,9 @@ if echo "$@" | grep -E -v -- '-c |-c=|--server-config |--server-config='; then
     SYS_PROPS+=" -c=standalone-ha.xml"
 fi
 
+# Adding support for JAVA_OPTS_APPEND
+sed -i '$a\\n# Append to JAVA_OPTS. Necessary to prevent some values being omitted if JAVA_OPTS is defined directly\nJAVA_OPTS=\"\$JAVA_OPTS \$JAVA_OPTS_APPEND\"' /opt/jboss/keycloak/bin/standalone.conf
```
slaskawi, line 108: This seems to be very fragile. Perhaps you could add this at the end of the file?

radudd: Thanks @slaskaw for reviewing. Adding support for this option will allow to add profile features to Keycloak https://www.keycloak.org/docs/latest/server_installation/#profiles. For my particular scenario, I would need to enable upload_script profile.

radudd: > This seems to be very fragile. Perhaps you could add this at the end of the file?
Yes, I agree. But the standalone.conf file is not part of the repo and is extracted from the Keycloak binary which is downloaded and installed during the build.

slaskawi: Ok, got it.

radudd:
Thanks @iankko and @slaskawi for the review and feedback. I just added an example on how to use JAVA_OPTS_APPEND

slaskawi (approved):
LGTM, let's give @iankko a chance to look into it.

iankko (approved):
/lgtm Thanks! @radudd 

(having the container image with the change built as with `0552f7ce2dc0`, and tried to enable
 e.g. the new account console via:

 ```
     $ docker run -it \
     -e KEYCLOAK_USER=admin \
     -e KEYCLOAK_PASSWORD=admin \
     -e JAVA_OPTS_APPEND="-Dkeycloak.profile=preview" \
     0552f7ce2dc0
````
the change works fine / as expected (`keycloak-preview` appears as configurable account theme &
once used / saved, its displayed when the user goes to account management).

slaskawi:
Integrated, thanks @radudd

PR #283, opened 2020-08-19
INTLY-8355 Better error logs on failed extension download

MStokluska:
## Description
This is a slight refactor of the extensions.sh script.

## Changes
Report extension download failure when the curl response code is different than 200. For example, if github.com is currently down and the extension can not be downloaded.

## Why
Previously, download failure was ignored and KeyCloak was started without the required plugin. 

Jira associated with the issue: https://issues.redhat.com/browse/INTLY-8355

### DIFF
--- a/keycloak-init-container/extensions.sh
+++ b/keycloak-init-container/extensions.sh
@@ -13,23 +13,42 @@ download_extension() {
     if [[ -z "$EXTENSION_URL" ]]; then
         return
     fi
-
+    echo 
     echo "Downloading extension from $EXTENSION_URL"
-    local FILENAME="$(curl --verbose --location --remote-name --remote-header-name --write-out '%{filename_effective}' --silent "$EXTENSION_URL" 2> /tmp/headers)"
+    local CURL_COMMAND="$(curl --verbose --location  --remote-name --remote-header-name --write-out "%{http_code} %{filename_effective}"  --silent "$EXTENSION_URL" 2> /tmp/headers)"
+  
+    local STATUS_CODE=${CURL_COMMAND:0:3}
 
-    # Try to get the filename from the response headers and return
-    # a random name if that fails
-    if ! grep -q -i '^< content-disposition:.*filename=' /tmp/headers ; then
-        local F="$(od -N8 -tx1 -An -v /dev/urandom | tr -d " \n").jar"
-        mv "$FILENAME" "$F"
-        FILENAME="$F"
+    if [ $STATUS_CODE -eq "200" ]; then
+        local FILENAME=${CURL_COMMAND:4}
+        echo "Extension downloaded successfully"
+
+        # Try to get the filename from the response headers and return
+        # a random name if that fails
+        if ! grep -q -i '^< content-disposition:.*filename=' /tmp/headers ; then
+            local F="$(od -N8 -tx1 -An -v /dev/urandom | tr -d "").jar"
+            mv "$FILENAME" "$F"
+            FILENAME="$F"
+        fi
+        echo " --> $FILENAME"
+    else
+        echo -e "Can not download the extension: $EXTENSION_URL\nError code: $STATUS_CODE"
+        ((STATUS+=1))
     fi
 
-    echo " --> $FILENAME"
 }
 
 # Parse the environment variable and download the extensions from the list
 IFS=,
-for EXT in $KEYCLOAK_EXTENSIONS ; do
+STATUS=0
+for EXT in ${KEYCLOAK_EXTENSION[@]} ; do
     download_extension "$EXT"
 done
+if [ "$STATUS" -ne 0 ]; then
+    echo
+    echo -e "Extensions.sh script failed at downloading all required extensions, number of failed downloads: $STATUS \n"
+    exit 1
+else
+    echo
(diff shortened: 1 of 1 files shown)


matskiv:
@mhajas @slaskawi Hey guys, this a bug fix related to extensions download mechanism for keyloak-operator. Could you review this whenever you have a minute?
Thank you!

mhajas (commented):
Thank you for the PR @MStokluska. I added a comment in the code.

Currently, when the first error is found we stop the execution, should we rather continue with downloading other extensions and then fail at the end? WDYT?

keycloak-init-container/extensions.sh
```diff
@@ -15,7 +15,18 @@ download_extension() {
     fi
 
     echo "Downloading extension from $EXTENSION_URL"
-    local FILENAME="$(curl --verbose --location --remote-name --remote-header-name --write-out '%{filename_effective}' --silent "$EXTENSION_URL" 2> /tmp/headers)"
+    local curl_command="$(curl --verbose --location  --remote-name --remote-header-name --write-out "%{response_code} %{filename_effective}"  --silent "$EXTENSION_URL" 2> /tmp/headers)"
```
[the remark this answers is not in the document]
MStokluska: Thanks for valuable feedback - I've updated the variable names.

MStokluska:
@mhajas Thank you for your comments. I think that we could allow for the download of other extensions before failing, however, in my opinion, I don't see a benefit in doing this. In my understanding, if we fail to download one of the extensions that we need, but let other extensions download and failing at the end, we would still be getting an error? 
I might be missing something, could you tell what would we be an advantage in doing so @mhajas?

MStokluska:
On the other hand @mhajas, it could be useful to know which extension did not download and why? It might help with the investigation of failures so I think I would go ahead and change the code to only fail in the end after running through download of all extensions if you think that is a better idea!

mhajas:
@MStokluska, I was just thinking aloud. Now I am not sure whether it is worth the effort, probably it is enough to just fail with the first, however, I would include the status code in the error message.

I have one more thing to think about, what will happen if it fails to download? It looks like the init-pod restartPolicy [will be set to](https://kubernetes.io/docs/concepts/workloads/pods/init-containers/#detailed-behavior) `onFailure` because restartPolicy of the keycloak pod is defaulting to `Always`. I was not able to find out how many times it will be tried before it gives up. Should we update https://github.com/keycloak/keycloak-operator/blob/master/pkg/model/init_container_common.go to restrict number of tries, or to set restartPolicy to `Never`? @slaskawi  WDYT?

slaskawi:
@MStokluska @mhajas I don't have a strong opinion on trying to download multiple plugins. I would probably go for the easiest thing to implement. But the error message needs to be clear. 

As for the retry policy, I would leave the default. Kubernetes uses a typical "eventual consistency" model. So the best thing to do is to throw an error in the init container and let Kubernetes scheduler does its work.

keycloak-init-container/extensions.sh
```diff
@@ -15,7 +15,18 @@ download_extension() {
...
     echo "Downloading extension from $EXTENSION_URL"
-    local FILENAME="$(curl --verbose --location --remote-name --remote-header-name --write-out '%{filename_effective}' --silent "$EXTENSION_URL" 2> /tmp/headers)"
+    local CURL_COMMAND="$(curl --verbose --location  --remote-name --remote-header-name --write-out "%{response_code} %{filename_effective}"  --silent "$EXTENSION_URL" 2> /tmp/headers)"
+
+    local STATUS_CODE=${CURL_COMMAND:0:3}
+
+    if [ $STATUS_CODE -eq "200" ]; then
+        echo "Extension downloaded successfully"
+    else
+        echo "Can not download the extension"
```
slaskawi, line 25: Could we return, what extension exactly was causing problems?

MStokluska: Thank you for your comment @slaskawi, sure, I'll update the PR and try to figure out the error codes, for now, any unsuccessful return is 000 while a successful is 200.

slaskawi: Also, watch out for curl errors... e.g. what happens if network is unavailable etc. 

The aim here is to give an administrator enough information to figure out whether it's local Kubernetes problems or maybe the target web page misbehaves (e.g. Github returning 503 etc).

MStokluska: Thanks @slaskawi , I'll give it a go!

MStokluska:
@slaskawi, @mhajas I've updated the PR so that now all extensions are being downloaded, and if an error occurs at one or more this the the output: 
image

The flow would be:
1. We are trying to download extension, if we get "200" status code we proceed with getting the filename from response headers `if` block.
2. If we fail we are displaying a name of extension that failed + status code
3. In the end we are displaying overall status: success or fail - if `fail` we are displaying the amount of failed downloads.

What do you guys think?

slaskawi (approved):
I love it! 

@mhajas any suggestions? If not, I'll merge it.

mhajas (approved):
@MStokluska Thanks for the changes, I like it. @slaskawi I think this is ready to merge.

slaskawi:
Integrated, thanks @MStokluska

PR #284, opened 2020-08-28
KEYCLOAK-15387 Add user and generate password for them, if user and password not passed

14DF397FCA:
Hello!
Small changes that allow to add default user and generate a password for them the user (administrator) did not pass it.

PS
I can't find how to create issue in your Jira.

### DIFF
--- a/server/README.md
+++ b/server/README.md
@@ -22,10 +22,23 @@ To be able to open Keycloak on localhost map port 8080 locally
 
 ## Creating admin account
 
-By default there is no admin user created so you won't be able to login to the admin console. To create an admin account
-you need to use environment variables to pass in an initial username and password. This is done by running:
+To be able to login to the admin console you need to have an admin user.  
+To create an admin account with your login and password you need to use environment variables to pass in an 
+initial username and password. This is done by running:
 
     docker run -e KEYCLOAK_USER=<USERNAME> -e KEYCLOAK_PASSWORD=<PASSWORD> jboss/keycloak
+
+When you run the container without the `KEYCLOAK_USER` and `KEYCLOAK_PASSWORD` variables, the username and password 
+of the admin user are generated automatically. You can find the actual username / password of the admin account with:
+    
+    docker logs <CONTAINER> | grep 'Username and password of the Keycloak admin account' -A3
+
+You will see output similar to the following one:
+    
+    docker logs  <CONTAINER> | grep 'Username and password of the Keycloak admin account' -A3
+        Username and password of the Keycloak admin account:
+        KEYCLOAK_USER: admin
+        KEYCLOAK_PASSWORD: 4FC1sqgf#hB1A/NuLn0
 
 You can also create an account on an already running container by running:
 
@@ -34,6 +47,8 @@ You can also create an account on an already running container by running:
 Then restarting the container:
 
     docker restart <CONTAINER>
+
+
 
 ### Providing the username and password via files
 
--- a/server/tools/docker-entrypoint.sh
+++ b/server/tools/docker-entrypoint.sh
@@ -28,6 +28,8 @@ file_env() {
 }
 
 SYS_PROPS=""
+DEFAULT_USER="admin"
+DEFAULT_PASSWORD=$(openssl rand -base64 20 | cut -c-20)
 
 ##################
 # Add admin user #
@@ -35,6 +37,9 @@ SYS_PROPS=""
 
 file_env 'KEYCLOAK_USER'
 file_env 'KEYCLOAK_PASSWORD'
+
+KEYCLOAK_USER=${KEYCLOAK_USER:-$DEFAULT_USER}
+KEYCLOAK_PASSWORD=${KEYCLOAK_PASSWORD:-$DEFAULT_PASSWORD}
 
 if [[ -n ${KEYCLOAK_USER:-} && -n ${KEYCLOAK_PASSWORD:-} ]]; then
     /opt/jboss/keycloak/bin/add-user-keycloak.sh --user "$KEYCLOAK_USER" --password "$KEYCLOAK_PASSWORD"
@@ -228,6 +233,19 @@ fi
 /opt/jboss/tools/statistics.sh
 /opt/jboss/tools/autorun.sh
 /opt/jboss/tools/vault.sh
+
+##################################
+# Print first login and password #
+##################################
+
+echo "========================================================================="
+echo ""
+echo "  Username and password of the Keycloak admin account:"
+echo "  KEYCLOAK_USER: ${KEYCLOAK_USER}"
+echo "  KEYCLOAK_PASSWORD: ${KEYCLOAK_PASSWORD}"
+echo ""
+echo "========================================================================="
+echo ""
 
 ##################
 # Start Keycloak #


slaskawi:
I like this feature! @stianst @abstractj @ASzc @iankko What do you think about this?

>  I can't find how to create issue in your Jira.

@14DF397FCA Please use navigate to https://issues.redhat.com/browse/KEYCLOAK, log in and hit "create" button (the blue one on the top bar).

slaskawi (changes requested):
The code LGTM but we'd need you to add a JIRA ID to the commit message and it would be great to have a short note about the new behavior somewhere in the documentation.

14DF397FCA:
Hello!
I updated documentation with new behaviour.

iankko:
TL; DR: +1 for adding this functionality (see below for longer reasoning).

@14DF397FCA The change looks reasonable (as very nice RFE) to me. It adds / follows the logic app should be runnable / launchable with as the very least possible a priori configuration necessary as possible. Besides that RH-SSO images have similar feature (administrator credentials are also auto-generated if not specified at the moment of image start, [1](https://github.com/jboss-container-images/redhat-sso-7-openshift-image/blob/sso74-dev/modules/sso/config/launch/setup/74/added/launch/add-sso-admin-user.sh#L12), [2](https://github.com/jboss-container-images/redhat-sso-7-openshift-image/blob/sso74-dev/templates/sso74-x509-https.json#L74)), which over the time of using the image has shown to be a very handy feature.

Just couple of very small documentation nitpicks. See upcoming comments in the docs changes below.

server/README.md
```diff
@@ -22,11 +22,24 @@ To be able to open Keycloak on localhost map port 8080 locally
 
 ## Creating admin account
 
-By default there is no admin user created so you won't be able to login to the admin console. To create an admin account
-you need to use environment variables to pass in an initial username and password. This is done by running:
+To be able to login to the admin console you need to have admin user.
```
iankko, line 25: (Nitpick) Replace with ```to have an admin user.```?

server/README.md
```diff
@@ -22,11 +22,24 @@ To be able to open Keycloak on localhost map port 8080 locally
...
 
-By default there is no admin user created so you won't be able to login to the admin console. To create an admin account
-you need to use environment variables to pass in an initial username and password. This is done by running:
+To be able to login to the admin console you need to have admin user.  
+To create an admin account with your login and password you need to use environment variables to pass in an 
+initial username and password. This is done by running:
 
     docker run -e KEYCLOAK_USER=<USERNAME> -e KEYCLOAK_PASSWORD=<PASSWORD> jboss/keycloak
 
+When you run the container without `KEYCLOAK_USER` and `KEYCLOAK_PASSWORD` user and password are generated
```
iankko, line 31: I think there should be a comma between the first part of the ```When``` sentence, and its continuation. Like:

<pre>
When you run the container without the `KEYCLOAK_USER` and `KEYCLOAK_PASSWORD` variables,
the username and password of the admin user are generated automatically.
</pre>

And I would end up the sentence after ```automatically``` (shorter sentences are better to read / understand).

server/README.md
```diff
@@ -22,11 +22,24 @@ To be able to open Keycloak on localhost map port 8080 locally
...
-By default there is no admin user created so you won't be able to login to the admin console. To create an admin account
-you need to use environment variables to pass in an initial username and password. This is done by running:
+To be able to login to the admin console you need to have admin user.  
+To create an admin account with your login and password you need to use environment variables to pass in an 
+initial username and password. This is done by running:
 
     docker run -e KEYCLOAK_USER=<USERNAME> -e KEYCLOAK_PASSWORD=<PASSWORD> jboss/keycloak
 
+When you run the container without `KEYCLOAK_USER` and `KEYCLOAK_PASSWORD` user and password are generated 
+automatically, you can find it with:
```
iankko, line 32: Start a new sentence, ```You can find the actual username / password of the admin account with:```. Yet, the code below should be modified to display the username / password credentials combination for both possible cases, how Keycloak can be started (IOW the ones specified via ```KEYCLOAK_USER``` / ```KEYCLOAK_PASSWORD``` combination, or the autogenerated ones).

server/README.md
```diff
@@ -22,11 +22,24 @@ To be able to open Keycloak on localhost map port 8080 locally
...
+    
+    docker logs <CONTAINER> | grep 'Your first login and password' -A3
+
+You will find next lines:
```
iankko, line 36: Replace ```You will find next lines:```above  with ```You will see output similar to the following one:```


(```You will find next lines:``` somehow evokes impression to me there will be ```Password: 4FC1sqgf#hB1A/NuLn0``` everytime, which is not the case)

server/tools/docker-entrypoint.sh
```diff
@@ -38,6 +39,9 @@ file_env 'KEYCLOAK_PASSWORD'
 
 if [[ -n ${KEYCLOAK_USER:-} && -n ${KEYCLOAK_PASSWORD:-} ]]; then
     /opt/jboss/keycloak/bin/add-user-keycloak.sh --user "$KEYCLOAK_USER" --password "$KEYCLOAK_PASSWORD"
+else
+    DEFAULT_PASSWORD=$(openssl rand -base64 20 | cut -c-20)
```
iankko, line 43: Move this ```DEFAULT_PASSWORD``` variable initialization above (at line 31, after the initialization of the ```DEFAULT_USER``` variable).

server/tools/docker-entrypoint.sh
```diff
@@ -38,6 +39,9 @@ file_env 'KEYCLOAK_PASSWORD'
 
 if [[ -n ${KEYCLOAK_USER:-} && -n ${KEYCLOAK_PASSWORD:-} ]]; then
     /opt/jboss/keycloak/bin/add-user-keycloak.sh --user "$KEYCLOAK_USER" --password "$KEYCLOAK_PASSWORD"
+else
```
iankko, line 42: You can drop this ```else``` branch and replace the empty string initialization in the previous ```if``` as follows (IOW simplify the whole if/else e.g. as):

```
KEYCLOAK_USER=${KEYCLOAK_USER:-$DEFAULT_USER}
KEYCLOAK_PASSWORD=${KEYCLOAK_PASSWORD:-$DEFAULT_PASSWORD}


if [[ -n ${KEYCLOAK_USER} && -n ${KEYCLOAK_PASSWORD} ]]; then
    /opt/jboss/keycloak/bin/add-user-keycloak.sh --user "$KEYCLOAK_USER" --password "$KEYCLOAK_PASSWORD"
```

This way, one one call to `add-user-keycloak.sh` would cover adding admin credentials in both cases (the specified ones, or the default ones) IIUC.

server/tools/docker-entrypoint.sh
```diff
@@ -38,6 +39,9 @@ file_env 'KEYCLOAK_PASSWORD'
...
+    DEFAULT_PASSWORD=$(openssl rand -base64 20 | cut -c-20)
+    /opt/jboss/keycloak/bin/add-user-keycloak.sh --user "${DEFAULT_USER}" --password "${DEFAULT_PASSWORD}"
```
iankko, line 44: Drop this line, if the `if/else` statement above is modified as suggested.

server/tools/docker-entrypoint.sh
```diff
@@ -229,6 +233,19 @@ fi
...
 /opt/jboss/tools/vault.sh
 
+##################################
+# Print first login and password #
+##################################
+
+echo "========================================================================="
+echo ""
+echo "  Your first login and password:"
+echo "  Login: ${DEFAULT_USER}"
```
iankko, line 243: Replace ```DEFAULT_USER``` / ```DEFAULT_PASSWORD``` below with their ```KEYCLOAK_USER``` / ```KEYCLOAK_PASSWORD``` counterparts. This way, it would just reprint the username / password credentials combination for the admin user, if the variables were specified. Or display the autogenerated / actual password.

server/tools/docker-entrypoint.sh
```diff
@@ -229,6 +233,19 @@ fi
 /opt/jboss/tools/autorun.sh
 /opt/jboss/tools/vault.sh
 
+##################################
+# Print first login and password #
+##################################
+
+echo "========================================================================="
+echo ""
+echo "  Your first login and password:"
```
iankko, line 242: Replace ```Your first login and password:``` with ```Username and password of the Keycloak admin account:```

iankko (changes requested):
The idea LGTM in general (as a whole). Just couple of nitpicks (both in the docs & code) to get the RFE polished

14DF397FCA:
@iankko , I updated my pull request with your notices.

server/README.md
```diff
@@ -22,23 +22,23 @@ To be able to open Keycloak on localhost map port 8080 locally
...
+When you run the container without the `KEYCLOAK_USER` and `KEYCLOAK_PASSWORD` variables, the username and password 
+of the admin user are generated automatically. You can find the actual username / password of the admin account with:
     
     docker logs <CONTAINER> | grep 'Your first login and password' -A3
 
-You will find next lines:
+You will see output similar to the following one:
     
     docker logs  <CONTAINER> | grep 'Your first login and password' -A3
         Your first login and password:
```
iankko, line 39: @14DF397FCA Thank you for your changes! 

Please replace ```Your first login and password:``` above in the docs yet with ```Username and password of the Keycloak admin account:``` so the docs is consistent with the code.

And replace ```KEYCLOAK_USER``` with ```Login:```, and ```KEYCLOAK_PASSWORD``` with ```Password:``` strings below, so it's again consistent with the code.

Sorry, if being a way too much pedantic :(

14DF397FCA: @iankko , I updated the request.

Don't worry)

iankko (approved):
@14DF397FCA Thank you for all the changes! The latest version LGTM. Please squash the four commits into one final one, and IMHO the final change is g2g

@slaskawi Please re-take a look at this once got a chance too (IMHO it's OK post commits are squashed)

slaskawi (changes requested):
The code itself LGTM.

Please squash all the commits (into 1), add JIRA ID and I'll merge it.

server/README.md
```diff
@@ -22,11 +22,24 @@ To be able to open Keycloak on localhost map port 8080 locally
...
+To be able to login to the admin console you need to have an admin user.  
+To create an admin account with your login and password you need to use environment variables to pass in an 
+initial username and password. This is done by running:
 
     docker run -e KEYCLOAK_USER=<USERNAME> -e KEYCLOAK_PASSWORD=<PASSWORD> jboss/keycloak
 
+When you run the container without the `KEYCLOAK_USER` and `KEYCLOAK_PASSWORD` variables, the username and password 
+of the admin user are generated automatically. You can find the actual username / password of the admin account with:
+    
+    docker logs <CONTAINER> | grep 'Your first login and password' -A3
```
iankko, line 34: @14DF397FCA Overlooked one issue yesterday yet :(. Please change the ```grep``` search pattern above to match the updated message. E.g. Replace the above command with:

```
docker logs <CONTAINER> | grep 'Username and password of the Keycloak admin account' -A3
```

server/README.md
```diff
@@ -22,11 +22,24 @@ To be able to open Keycloak on localhost map port 8080 locally
...
+
+You will see output similar to the following one:
+    
+    docker logs  <CONTAINER> | grep 'Your first login and password' -A3
```
iankko, line 38: And please here too. Thanks!

iankko (changes requested):
Two docs changes are needed yet (see comments above).

14DF397FCA:
@slaskawi I squashed commits, Jira ID to this pull request. Is the pull request ok now?

iankko:
> @slaskawi I squashed commits, Jira ID to this pull request. Is the pull request ok now?

@14DF397FCA Please list the JIRA ID in both the PR label (already present), and also in the git commit message (please add it yet).

iankko (approved):
/lgtm

14DF397FCA:
> > @slaskawi I squashed commits, Jira ID to this pull request. Is the pull request ok now?
> 
> @14DF397FCA Please list the JIRA ID in both the PR label (already present), and also in the git commit message (please add it yet).

@iankko I updated commit's title.

iankko:
@abstractj 

> @slaskawi I believe worth a discussion on the dev mailing list. If we include it, we make the username predictable, which means increasing the attack surface now that we have a predictable username. Another thing that's critical from my point of view is the fact that we can see the username/password at the logs. We should never record it.

+1 FWIW discussing this on the ML. WRT to predictable username, it can be made random too (this is what Red Hat Single Sign-On on OpenJDK / OpenJ9 images are actually using thx to the templates:
* [SSO_ADMIN_USERNAME auto-generation](https://github.com/jboss-container-images/redhat-sso-7-openshift-image/blob/sso74-dev/templates/sso74-x509-https.json#L74)
* [SSO_ADMIN_PASSWORD auto-generation](https://github.com/jboss-container-images/redhat-sso-7-openshift-image/blob/sso74-dev/templates/sso74-x509-https.json#L82)

> 
> If we really would like to move forward with this idea, my suggestion is:
> 
>     1. Make username/password random. Something like Docker does by naming its images.

+1 (not only Docker, but also our product images do the same)

> 
>     2. Print it, but never stored at the logs

You mean to use different filedescriptor than the logging facility itself to print out the actual credentials combination that got auto-generated?

> 
>     3. It should not be the default behavior, but instead a flag which people explicitly tell to generate the username/password

For product images OpenShift ensures (when templates are used) both variables are auto-generated (so its tricky to guess). For Keycloak images it's a bit tricky, since the image itself doesn't need to be run solely on OpenShift (we need to support also Docker, Kubernetes, etc.)

> 
> 
> Being very honest here, adding this change creates some inconsistencies with the Keycloak server. Today we do not provide a default username, nor generate passwords. But that's just my personal opinion and keycloak-dev should be a better place to reach a wider audience.

Fair enough, thx for sharing!

@14DF397FCA PTAL -^ Could you please submit a post to the keycloak-dev mailing list describing the intended change, so it's possible to obtain wider feedback on how this change should be implemented?

Thank you!
Jan

stianst:
This can't be added at least not in it's current form. If someone starts a DB, then starts one container with KEYCLOAK_USERNAME=superuser, then starts another container without KEYCLOAK_USERNAME they would end up with two admin accounts (superuser, and admin).

stianst:
I'm not really in favour of adding this. Not because I have a big issue with generating username/password, but more from the fact that I think the current approach that this extends is a bit broken.

slaskawi:
> This can't be added at least not in it's current form. If someone starts a DB, then starts one container with KEYCLOAK_USERNAME=superuser, then starts another container without KEYCLOAK_USERNAME they would end up with two admin accounts (superuser, and admin).

@stianst In this case you're assuming multiple Keycloak containers, with different configurations pointing to the same database. I'm not sure if this is a valid use case. All Keycloak instances should always have the same configuration.

But this touches an interesting topic, when we invoke the add user script, we don't check if an admin account has already been created. I recently noticed that in Keycloak Operator. Perhaps we should?

slaskawi:
Let me close this for now @14DF397FCA 

Please create a discussion on [Keycloak DEV](https://groups.google.com/forum/#!forum/keycloak-dev) and once we get an agreement whether this should go in or not, please reopen this if needed.

PR #285, opened 2020-09-03
Handle databases set via CloudFoundry VCAP_SERVICES

dwaynebailey:
* Extract credentials for a datbase names DB_VCAP_NAME
* Use grep to ensure we use existing tools on the current image without
  loading the image with a new tool like jq.
* DB_VENDOR is still set externally so that the script doesn't have to
  guess it correctly

Sorry haven't quite followed the guidelines on contributing.  Just wanted to push an initial PR for comment about my approach.

### DIFF
--- a/server/tools/docker-entrypoint.sh
+++ b/server/tools/docker-entrypoint.sh
@@ -153,6 +153,16 @@ if [[ -z ${DB_VENDOR:-} ]]; then
     export DB_VENDOR="h2"
 fi
 
+# Handle VCAP_SERVICES databases
+if [[ ! -z ${DB_VCAP_NAME:-} ]]; then
+    vcap_credentials=$(echo $VCAP_SERVICES | grep -Po '(?<="name": "'${DB_VCAP_NAME}'").*}' | grep -Po '(?<="credentials": { )[^}]*')
+    export DB_ADDR=$(echo $vcap_credentials | grep -Po '(?<="host": ")[^"]+')
+    export DB_PORT=$(echo $vcap_credentials | grep -Po '(?<="port": )\d+')
+    export DB_DATABASE=$(echo $vcap_credentials | grep -Po '(?<="name": ")[^"]+')
+    export DB_USER=$(echo $vcap_credentials | grep -Po '(?<="username": ")[^"]+')
+    export DB_PASSWORD=$(echo $vcap_credentials | grep -Po '(?<="password": ")[^"]+')
+fi
+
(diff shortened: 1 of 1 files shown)


stianst:
I'm not too keen on adding CloudFoundry specific things to the container image. Is this something that images on Docker Hub / Quay.io supports? If so please point to other popular images that support this.

Also, please provide some more details on what the use-case is here.

dwaynebailey:
I don't use docker images much on CloudFoundry so not 100% sure how others have approached this.

ito use case, this is an outline of the issue.

In CloudFoundry I will:
1. create a DB
2. create an app manifest that specifies that the app has access to the named DB
3. Deploy the app
4. At deployment time the binding between app and DB occurs and the database credential are exposed in the `VCAP_SERVICES` environment variable from where they can be extracted by the app.

Summary is this.  I don't have the settings for the DB until I actually deploy the docker image as an app.

There are ways around this, which I currently do, which is to:
1. Create the DB
2. Create a fake empty docker app
3. Bind the two
4. Extract the DB credentials
5. Redeploy the docker app with the DB credentials that we now know.

This makes deployment a lot more complicated ito manual processes or automation.  And has a few gotchas when I'm doing deployments via Terraform.

Trust that helps.

slaskawi:
@dwaynebailey Could you please double check the documentation to see if there's any way to perform from `VCAP_SERVICES` env. variable to some others? I believe there should be a way to do that (I would will be very surprised if there's none).

slaskawi:
@dwaynebailey May I ask you to look at my previous comment?

dwaynebailey:
@slaskawi do you mean that some other env variables contain the data you'd find in `VCAP_SERVICES`? Simple answer would be no the connection data is only in that variable.   `VCAP_SERVICES` is a way to convey credentials for many types of bound services, not just a database see https://docs.cloudfoundry.org/devguide/deploy-apps/environment-variable.html#VCAP-SERVICES

stianst (changes requested):
I'm not in favour of adding this as it is bespoke to CloudFoundry. We also have no way of testing/verifying if this actually works.

If you can show similar features added to other containers, then we may reconsider, but in general we won't add CloudFoundry specfic capabilities to the container.

Another option is of course to extend the image perhaps.

slaskawi:
I agree with @stianst here. 

Let me close it for now. @dwaynebailey if you find a generic solution how other community projects deal with it, I'm happy to reconsider.

PR #287, opened 2020-09-08
Update of docker-entrypoint to feature upload_scripts

netoralves:


stianst:
Thanks for the PR, but I'm afraid we need a more generic way to enable/disable features than this. One thing to point out here as well is that upload-scripts is a deprecated feature (for security reasons) that will soon be removed.

PR #289, opened 2020-09-14
KEYCLOAK-15541 Keycloak docker image fails to connect to MSSQL when p…

antaran13:
KEYCLOAK-15541 Keycloak docker image fails to connect to MSSQL when password contains special characters

JIRA issue: https://issues.redhat.com/browse/KEYCLOAK-15541
Mailing list: https://groups.google.com/g/keycloak-dev/c/qh7PG754ThQ

It seems to me that the simplest way how to avoid exception:

```
Caused by: com.microsoft.sqlserver.jdbc.SQLServerException: The connection string contains a badly formed name or value.
	at com.microsoft.sqlserver.jdbc//com.microsoft.sqlserver.jdbc.SQLServerException.makeFromDriverError(SQLServerException.java:234)
```
when password is like "password}{" is to just remove username and password from JDBC string all together.

### DIFF
--- a/server/tools/cli/databases/mssql/change-database.cli
+++ b/server/tools/cli/databases/mssql/change-database.cli
@@ -1,5 +1,5 @@
 /subsystem=datasources/data-source=KeycloakDS: remove()
-/subsystem=datasources/data-source=KeycloakDS: add(jndi-name=java:jboss/datasources/KeycloakDS,enabled=true,use-java-context=true,use-ccm=true, connection-url="jdbc:sqlserver://${env.DB_ADDR:mssql}:${env.DB_PORT:1433};databaseName=${env.DB_DATABASE:keycloak};sendStringParametersAsUnicode=false;user=${env.DB_USER:keycloak};password=${env.DB_PASSWORD:password};${env.JDBC_PARAMS:}", driver-name=sqlserver)
+/subsystem=datasources/data-source=KeycloakDS: add(jndi-name=java:jboss/datasources/KeycloakDS,enabled=true,use-java-context=true,use-ccm=true, connection-url="jdbc:sqlserver://${env.DB_ADDR:mssql}:${env.DB_PORT:1433};databaseName=${env.DB_DATABASE:keycloak};sendStringParametersAsUnicode=false;${env.JDBC_PARAMS:}", driver-name=sqlserver)
 /subsystem=datasources/data-source=KeycloakDS: write-attribute(name=user-name, value=${env.DB_USER:keycloak})
 /subsystem=datasources/data-source=KeycloakDS: write-attribute(name=password, value=${env.DB_PASSWORD:password})
 /subsystem=datasources/data-source=KeycloakDS: write-attribute(name=check-valid-connection-sql, value="SELECT 1")


server/tools/cli/databases/mssql/change-database.cli
```diff
@@ -1,5 +1,5 @@
 /subsystem=datasources/data-source=KeycloakDS: remove()
-/subsystem=datasources/data-source=KeycloakDS: add(jndi-name=java:jboss/datasources/KeycloakDS,enabled=true,use-java-context=true,use-ccm=true, connection-url="jdbc:sqlserver://${env.DB_ADDR:mssql}:${env.DB_PORT:1433};databaseName=${env.DB_DATABASE:keycloak};sendStringParametersAsUnicode=false;user=${env.DB_USER:keycloak};password=${env.DB_PASSWORD:password};${env.JDBC_PARAMS:}", driver-name=sqlserver)
+/subsystem=datasources/data-source=KeycloakDS: add(jndi-name=java:jboss/datasources/KeycloakDS,enabled=true,use-java-context=true,use-ccm=true, connection-url="jdbc:sqlserver://${env.DB_ADDR:mssql}:${env.DB_PORT:1433};databaseName=${env.DB_DATABASE:keycloak};sendStringParametersAsUnicode=false;${env.JDBC_PARAMS:}", driver-name=sqlserver)
```
slaskawi, line 2: This seems to me like a non-backwards compatible change. If someone used `env.DB_USER` and `env.DB_PASSWORD`, this PR would break his code, wouldn't it?

antaran13: Please have a look at line 3 and 4, where username and password are set directly.

`/subsystem=datasources/data-source=KeycloakDS: write-attribute(name=user-name, value=${env.DB_USER:keycloak})
/subsystem=datasources/data-source=KeycloakDS: write-attribute(name=password, value=${env.DB_PASSWORD:password})`

It does not really matter even if you constructed your URL with dummy username and password like so:

`/subsystem=datasources/data-source=KeycloakDS: add(jndi-name=java:jboss/datasources/KeycloakDS,enabled=true,use-java-context=true,use-ccm=true, connection-url="jdbc:sqlserver://${env.DB_ADDR:mssql}:${env.DB_PORT:1433};databaseName=${env.DB_DATABASE:keycloak};sendStringParametersAsUnicode=false;user=dummy;password=dummy;${env.JDBC_PARAMS:}", driver-name=sqlserver)`

It would still work and be backward compatible, because username and password set on line 3 and 4 override whatever username and password was provided by URL. As far as know having username and password in URL and set it directly as well does not serve any purpose, it is just confusing.

antaran13: This change is consistent with how `change-database.cli` scripts for all other databases work.
I have checked all of them and none of them provides username and password in connection url.

In case you are in doubt you can build it locally and try to run `keycloak-mssql.yml` from `docker-compose-examples`... it should work as before.

The only difference being, that now you should be able to use passwords which contain characters like "}{".
(There are 3 places where you need to change password in `keycloak-mssql.yml`.)

slaskawi: Ok, thanks for the explanation.

slaskawi (approved):
LGTM. Let's give @hmlnarik @mhajas @ASzc @miquelsi @abstractj and @vmuzikar a chance to look into it. I'll wait a few days and then merge it.

vmuzikar (commented):
@antaran13 Thank you for your contribution!
The change looks good as far as I can tell. I just wonder about tests. Do we have any tests for this which would make sense to extend to cover this?

slaskawi:
@vmuzikar Unfortunately we don't have any (direct) tests for this repo. We test some of the use cases in the Operator but it would be great to come up with some plan around designing a testsuite for both existing Keycloak Container as well as Quarkus-based distribution. Perhaps @pedroigor could also give us some hints on testing?

vmuzikar:
I'm not sure but don't we have some testing pipeline for images? @iankko?

iankko:
> I'm not sure but don't we have some testing pipeline for images? @iankko?

@vmuzikar No automated testing (pipeline) available for the community images yet, sorry :(. It would need to be designed.

slaskawi:
Integrated, thanks @antaran13

PR #292, opened 2020-10-09
[KEYCLOAK-15899] Set default-provider value for spi vault

adrien-dorsaz:
According to the [documentation](https://www.keycloak.org/docs/latest/server_admin/index.html#kubernetes-openshift-files-plaintext-vault-provider) the `<spi>` has the `<default-provider>` attribute to select the provider for the vault.

That value seems required to be able to use the vault with OpenShift.

### DIFF
--- a/server/tools/cli/files-plaintext-vault.cli
+++ b/server/tools/cli/files-plaintext-vault.cli
@@ -2,5 +2,6 @@ embed-server --server-config=$configuration_file --std-out=discard
 echo ** Adding vault spi **
 /subsystem=keycloak-server/spi=vault/:add
 /subsystem=keycloak-server/spi=vault/provider=files-plaintext/:add(enabled=true,properties={dir => $plaintext_vault_provider_dir})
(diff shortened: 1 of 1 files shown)


slaskawi (approved):
LGTM, @hmlnarik @martin-kanis would you like to look into it?

hmlnarik (approved):

martin-kanis:
@adrien-dorsaz Thanks, it looks good. I think the command should be also added to documentation to match the script. Would you mind to add it here https://github.com/keycloak/keycloak-documentation/blob/master/server_admin/topics/vault.adoc and link the PR to the Jira?

adrien-dorsaz:
You are welcome, thanks for the review.

Indeed, that's better to update the documentation too, I've made the PR keycloak/keycloak-documentation#1048 and updated the JIRA to refer to this PR too.

martin-kanis (approved):

slaskawi:
Integrated, thanks @adrien-dorsaz

PR #295, opened 2020-10-15
[KEYCLOAK-15931] Add support for H2 server connections

tdiesler:


slaskawi (approved):
LGTM! Let's wait a few days for others to have a look at this.

miquelsi (approved):

ASzc (approved):
Seems like it would work

server/tools/docker-entrypoint.sh
```diff
@@ -145,6 +148,9 @@ if [[ -z ${DB_VENDOR:-} ]]; then
         export DB_VENDOR="oracle"
     elif (printenv | grep '^MSSQL_ADDR=' &>/dev/null); then
         export DB_VENDOR="mssql"
+    elif (printenv | grep '^H2_ADDR=' &>/dev/null); then
+        export DB_VENDOR="h2"
+        export DB_ADDR="h2"
```
vmuzikar, line 153: I'm not very familiar how this works exactly so I've got maybe a stupid question. :) Is it ok to set (or perhaps even override?) `DB_ADDR` here? Other DBs don't do that. What if someone wants to use a different address?

vmuzikar (commented):
@tdiesler Thank you for your contribution. As far as I can tell (I'm no expert in this area) looks good to me, although I haven't tested it.
Added one comment.

server/tools/docker-entrypoint.sh
tdiesler: In line 133 we check whether there is a host 'h2' on the network. This would be the case in k8s with a pod named 'h2' or on a custom docker network with a container named 'h2'. If neither is true, DB_VENDOR, DB_ADDR are not set.

Line 151, which you reference here is only there for completeness. It deals with legacy env vars, which will very likely never have H2_ADDR

For H2, unlike for the dbs, it is not enough to set the DB_VENDOR, because that would simply use the H2 embedded/in-memory database. Yes, if your H2 pod/container is called something different (e.g. h2db) , the DB_VENDOR defaults to 'h2' (as before) and you can set DB_ADDR=h2db which will then be used in change-database.cli

vmuzikar: Ok, makes sense I think. Thanks for clarifying.

vmuzikar (approved):

server/tools/docker-entrypoint.sh
tdiesler: Like this

```
# Keycloak Pod ########################################################################################################
---
apiVersion: v1
kind: Pod
metadata:
  name: keycloak
  labels:
    app: nessus
    pod: keycloak
spec:
  containers:
  - name: keycloak
    image: nessusio/keycloak
    env:
    - name: DB_VENDOR
      value: h2
    - name: DB_ADDR
      value: $(H2_HTTP_SERVICE_HOST)
    - name: DB_DATABASE
      value: fuse
    - name: DB_SCHEMA
      value: keycloak
    - name: KEYCLOAK_USER
      value: admin
    - name: KEYCLOAK_PASSWORD
      value: admin
    - name: KEYCLOAK_IMPORT
      value: https://raw.githubusercontent.com/tdiesler/nessus-actions/master/docs/k8s/deployment/keycloak/myrealm.json
    ports:
    - containerPort: 8080
    - containerPort: 8443
    volumeMounts:
    - name: tls
      mountPath: /etc/x509/https
  volumes:
  - name: tls
    secret: 
      secretName: keycloak
```

slaskawi:
Integrated. Thanks @tdiesler !

PR #301, opened 2020-11-12
KEYCLOAK-15886 Add health probe scripts to quarkus container

ASzc:
These scripts just call the smallrye health endpoints via curl. Nothing complicated.

### DIFF
--- a/server-x/probes/liveness.sh
+++ b/server-x/probes/liveness.sh
@@ -1,0 +1,2 @@
+#!/bin/bash -e
+curl -sSfo /dev/null --max-time 10 "http://localhost:8080/health/live"
--- a/server-x/Dockerfile
+++ b/server-x/Dockerfile
@@ -14,6 +14,7 @@ ADD tools /opt/jboss/tools
 FROM registry.access.redhat.com/ubi8-minimal
 
 COPY --from=build-env /opt/jboss /opt/jboss
+COPY probes /probes
 
 RUN microdnf update -y && \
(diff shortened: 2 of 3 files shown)


slaskawi (approved):
LGTM @ASzc !

@miquelsi @pedroigor I'll give you a few days just in case you wanted to review this changes.

server-x/probes/liveness.sh
```diff
@@ -0,0 +1,2 @@
+#!/bin/bash -e
+curl -sSfo /dev/null --max-time 10 "http://localhost:8080/auth/health/live"
```
miquelsi, line 2: The Quarkus distribution is placing Keycloak directly at the root of the URL, so no `/auth` context is needed here. The same for readiness check.

ASzc: Ok. I'll fix that. Updating [the spec](https://github.com/keycloak/keycloak-community/pull/55/files) would be a good idea.

miquelsi: Thanks, I've added a comment to the spec PR. Approving.

miquelsi (approved):

slaskawi (approved):
LGTM!

slaskawi:
Let's give @pedroigor a day or two to look at this. Then, I'll merge it.

pedroigor (approved):

slaskawi:
Integrated, thanks @ASzc !

PR #304, opened 2020-11-23
workaround for https://issues.redhat.com/browse/KEYCLOAK-16436  priva…

atlasloewenherz:
- if the tls.key private key is encrypted, the container skips it and create a new certificate
- this is a minor work around to get the password from **tls.pass** file containing the password that can be mounted as tls.crt, tls.key


atlasloewenherz:
Please do not merge yet!, need more testing still

slaskawi (commented):
Thanks for looking into this!

I think a slightly better way would be to mount password-protected tls keys and passwords in some other directory (maybe `/etc/x509/https/password-protected`?). Then, during the bootstrap process, one of the scripts would simply remove the password and put them into `/etc/x509/https` directory. This way your implementation would be more modular and the chances of breaking anything would be very low.

PR #308, opened 2020-12-08
Add env for specific jboss port offset

zeoe:
Within our Docker network, we do not want to run the keycloak on port 8080. For this reason I have created a new environment variable "KEYCLOAK_JBOSS_PORT_OFFSET". The variable passes the offset to jboss.socket.binding.port-offset and so the port can be changed

Thanks for looking over it!


ASzc:
Thanks for the PR!

I don't think that's going to work as you expect due to [EXPOSE being hardcoded in the image](https://docs.docker.com/engine/reference/builder/#expose). The server image has 8080 and 8443 exposed [here](https://github.com/keycloak/keycloak-containers/blob/master/server/Dockerfile#L27), and no other ports can be used.

You should already be able to do what you want; the port inside a container doesn't matter at all, except as a default. It's fully rebindable at container runtime by the `-p` switch. See `docker run` [documentation here](https://docs.docker.com/engine/reference/commandline/run/#publish-or-expose-port--p---expose).

zeoe:
Thank you for your quick response.

we checked it and yes we find a solution. Thank you very much for the hint!
I will close the PR.

PR #312, opened 2020-12-20
Fix keycloak version to 12.0.1 for server and server-x

Leletir:
Hello,

I was trying to build the server-x docker image but the version on the Dockerfile was 11.0.0 and the url was false, i just made this fixes and upgraded the Keycloak version to 12.0.1.

I don't know what i the right workflow to submit this PR, I'll be more than happy to make required changes to follow your guidelines.

Have a good day.


server-x/Dockerfile
```diff
@@ -1,7 +1,7 @@
 FROM registry.access.redhat.com/ubi8-minimal AS build-env
 
-ARG KEYCLOAK_VERSION=11.0.0
-ARG KEYCLOAK_DIST=https://github.com/keycloak/keycloak/releases/download/$KEYCLOAK_VERSION/keycloak.x-$KEYCLOAK_VERSION.tar.gz
+ARG KEYCLOAK_VERSION=12.0.1
```
slaskawi, line 3: Even though this is the correct fix, we also need to update the `set-version.sh` script to prevent this from happening again.

@Leletir Would you mind to add those changes?

slaskawi (changes requested):

server-x/Dockerfile
tanvp112: docker pull quay.io/keycloak/keycloak-x

Using default tag: latest
Error response from daemon: manifest for quay.io/keycloak/keycloak-x:latest not found: manifest unknown: manifest unknown

Checked https://quay.io/repository/keycloak/keycloak-x?tab=tags turns out no no image was tagged.

Is this related?

Leletir: Hi @slaskawi, thank for your feedback, indeed I missed it. It's fixed :)

slaskawi: Yes, looking at the Quay build logs, it seems the build was triggered but failed.

slaskawi (approved):
LGTM, let's give @stianst and @pedroigor a chance to review this.

pedroigor (approved):

Leletir:
Thanks @slaskawi @pedroigor for the review. Do you know what we have to do to mkae the keycloak-x image available ?

slaskawi:
The image is already available at
https://quay.io/repository/keycloak/keycloak-x?tab=tags

Leletir:
Hi @slaskawi, thank you for the link, I did not find it ...

By the way thank you guys fo the review, have a good day !

PR #315, opened 2021-03-11
[KEYCLOAK-17402] documents the need of Headless Service for OpenShiftand apply it to templates

ghost:


vmuzikar (changes requested):
@adrien-dorsaz Thank you for your contribution. Although it is a valid issue (I believe it's actually a bug), IMHO the solution to use the "main" service in headless mode as it disables load balancing. A more proper solution would be to create a separate discovery service just for this purpose, similarly what [the Operator does](https://github.com/keycloak/keycloak-operator/blob/master/pkg/model/keycloak_discovery_service.go).

adrien-dorsaz:
Thanks for the review.

> IMHO the solution to use the "main" service in headless mode is not correct as it disables load balancing.

Indeed, the Kubernetes load balancing is disabled and I thought it wasn't an issue, because I expected Keycloak doing itself the load balancing job.

Indeed I certainly misunderstood the DNS.ping feature.

If I understand correctly now, this feature requires real IP address of Keycloak containers, so they can talk together directly.

Although, it's not linked at all to load balancing, but it's used to share session state, config update... between containers, isn't it ?

>  A more proper solution would be to create a separate discovery service just for this purpose, similarly what the Operator does.

Thanks for the links of the Operator configuration, indeed, that's clever to create a separated discovery service, so we can keep the Kubernetes load balancing running.

adrien-dorsaz:
I've just updated the Pull Request to rebase it from master and to create a dedicated service for the discovery system.

stianst:

PR #316, opened 2021-04-01
extensions script cleanup

apeschel:
* Set pipefail
* Explicitly define all external variables used in this script
* Remove extraneous whitespace
* Correct shellcheck issues
* Consistently use shell builtins for logic
* Lower case for local variable names
* Better variable name for failed downloads

### DIFF
--- a/keycloak-init-container/extensions.sh
+++ b/keycloak-init-container/extensions.sh
@@ -1,52 +1,66 @@
-#!/bin/bash -e
+#!/bin/bash
 
-EXTENSIONS_VOLUME="${EXTENSIONS_VOLUME:-/opt/extensions}"
+set -o pipefail -o errexit -o nounset
 
-mkdir -p "$EXTENSIONS_VOLUME"
-cd "$EXTENSIONS_VOLUME"
-echo "Target directory: $EXTENSIONS_VOLUME"
+## External variables used by this script
+declare -rx KEYCLOAK_EXTENSIONS
+declare -rx EXTENSIONS_VOLUME="${EXTENSIONS_VOLUME:-/opt/extensions}"
 
 # Download a single extension
 download_extension() {
-    local EXTENSION_URL="$1"
+    local extension_url="$1"
 
-    if [[ -z "$EXTENSION_URL" ]]; then
+    if [[ -z "$extension_url" ]]; then
         return
     fi
-    echo 
-    echo "Downloading extension from $EXTENSION_URL"
-    local CURL_COMMAND="$(curl --verbose --location  --remote-name --remote-header-name --write-out "%{http_code} %{filename_effective}"  --silent "$EXTENSION_URL" 2> /tmp/headers)"
-  
-    local STATUS_CODE=${CURL_COMMAND:0:3}
+    echo
+    echo "Downloading extension from $extension_url"
+    local curl_command
+    curl_command="$(curl --verbose --location --remote-name --remote-header-name --write-out "%{http_code} %{filename_effective}" --silent "$extension_url" 2> /tmp/headers)"
 
-    if [ $STATUS_CODE -eq "200" ]; then
-        local FILENAME=${CURL_COMMAND:4}
+    local status_code=${curl_command:0:3}
+
+    if [[ "$status_code" -eq "200" ]]; then
+        local filename=${curl_command:4}
         echo "Extension downloaded successfully"
 
         # Try to get the filename from the response headers and return
         # a random name if that fails
-        if ! grep -q -i '^< content-disposition:.*filename=' /tmp/headers ; then
-            local F="$(od -N8 -tx1 -An -v /dev/urandom | tr -d "").jar"
-            mv "$FILENAME" "$F"
-            FILENAME="$F"
+        if ! grep -q -i '^< content-disposition:.*filename=' /tmp/headers; then
+            local F
+            F="$(od -N8 -tx1 -An -v /dev/urandom | tr -d "").jar"
+            mv "$filename" "$F"
+            filename="$F"
         fi
-        echo " --> $FILENAME"
+        echo " --> $filename"
     else
-        echo -e "Can not download the extension: $EXTENSION_URL\nError code: $STATUS_CODE"
-        ((STATUS+=1))
+        echo -e "Can not download the extension: ${extension_url}\nError code: ${status_code}"
+        ((FAILED_DOWNLOADS+=1))
+        return 1
     fi
-
 }
+
+mkdir -p "$EXTENSIONS_VOLUME"
+if ! cd "$EXTENSIONS_VOLUME"; then
+    echo "Failed to change directory to ${EXTENSIONS_VOLUME}"
+    exit 1
+fi
+echo "Target directory: ${EXTENSIONS_VOLUME}"
 
 # Parse the environment variable and download the extensions from the list
 IFS=,
-STATUS=0
-for EXT in ${KEYCLOAK_EXTENSIONS[@]} ; do
-    download_extension "$EXT"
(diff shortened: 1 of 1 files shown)


apeschel:
Some cleanup I did while debugging this script for https://github.com/keycloak/keycloak-operator/pull/340

On a side note, it appears `KEYCLOAK_EXTENSIONS` was switched to `KEYCLOAK_EXTENSION` accidentally in the past:

keycloak/keycloak-containers@da4b38c#diff-51b1ab20d6e750a4464ec80bb5021a47be580d89ccc4cd25427ee812c5893347L33-R44

vmuzikar:
@apeschel Thank you for your commit. Could you please create a JIRA and squash your commits? See the [contribution guide](https://github.com/keycloak/keycloak-containers/blob/master/CONTRIBUTING.md).

vmuzikar:
As this is (among other changes) fixes a critical issue, created a JIRA for it: https://issues.redhat.com/browse/KEYCLOAK-18143

vmuzikar:
@iankko @drichtarik Could you please review?

keycloak-init-container/extensions.sh
```diff
@@ -1,52 +1,63 @@
-#!/bin/bash -e
+#!/bin/bash
 
-EXTENSIONS_VOLUME="${EXTENSIONS_VOLUME:-/opt/extensions}"
+shopt -so errexit pipefail
```
iankko, line 3: @apeschel 

In which way the meaning of the previous `shopt` commad differs from `set -xeuo pipefail`? Did `shellcheck` suggest this form for you?

Or if the display of expanded `PS4` value isn't desired, could `set -euo pipefail` form be used? The latter form looks a more commonly used one for me, but that's just a nuance.

keycloak-init-container/extensions.sh
```diff
@@ -1,52 +1,63 @@
...
+            F="$(od -N8 -tx1 -An -v /dev/urandom | tr -d "").jar"
+            mv "$filename" "$F"
+            filename="$F"
         fi
-        echo " --> $FILENAME"
+        echo " --> $filename"
     else
-        echo -e "Can not download the extension: $EXTENSION_URL\nError code: $STATUS_CODE"
-        ((STATUS+=1))
+        echo -e "Can not download the extension: $extension_url\nError code: $status_code"
```
iankko, line 37: ... `$extension_url\nError code:`...

Would the above better to be enclosed in curly brackets, IOW ... `${extension_url}\nError code:`... (unsure if `\n` character is enough Bash not to try to expand the `$extension_url\nError` variable instead. Maybe it is).

keycloak-init-container/extensions.sh
```diff
@@ -1,52 +1,63 @@
...
+        ((FAILED_DOWNLOADS+=1))
```
iankko, line 38: Since changing variable names to lowercase, should this one be lowercased too?

keycloak-init-container/extensions.sh
```diff
@@ -1,52 +1,63 @@
...
+        return 1
     fi
-
 }
 
+mkdir -p "$EXTENSIONS_VOLUME"
+if ! cd "$EXTENSIONS_VOLUME"; then
+    echo "Cannot cd to $EXTENSIONS_VOLUME"
```
iankko, line 45: `s/"Cannot cd to $EXTENSIONS_VOLUME"/"Failed to change directory to $EXTENSIONS_VOLUME"/g`?

(error message shouldn't contain additional script commands / statements, but rather explain the reason in human language. But maybe again just issue of taste / nuance).

keycloak-init-container/extensions.sh
```diff
@@ -1,52 +1,63 @@
...
+echo "Target directory: $EXTENSIONS_VOLUME"
+
 # Parse the environment variable and download the extensions from the list
 IFS=,
-STATUS=0
-for EXT in ${KEYCLOAK_EXTENSION[@]} ; do
-    download_extension "$EXT"
+FAILED_DOWNLOADS=0
+
+for EXT in ${KEYCLOAK_EXTENSIONS[*]}; do
```
iankko, line 54: This is where `set -u` above would help to catch this kind of variable names typo.

iankko (commented):
/LGTM Thanks, @apeschel. Just couple of comments / nuances to clarify yet.

keycloak-init-container/extensions.sh
apeschel: @iankko `shopt` is a personal preference/style thing - I just find shopt has a cleaner interface than set.

I'll switch it to `set` since that appears to be the local preference.

apeschel: @iankko I'm curious where the `set -xeuo pipefail` comes from though -- there was just a `-e` on the shebang line previously.

keycloak-init-container/extensions.sh
apeschel: FAILED_DOWNLOADS is being used as an environment variable, and so should be capitalized -- I'll add an explicit export to make it clearer.

apeschel: It's being used as a global variable, which I typically leave uppercase -- would you prefer it to be lowercase?

apeschel:
@iankko Thanks for the review! I've been pretty busy so it took a while to get to following up. I made all your suggested changes, with the exception of the `FAILED_DOWNLOADS` casing, since it's a global variable. Let me know if you would still prefer it to be lowercase.

keycloak-init-container/extensions.sh
iankko: > @iankko I'm curious where the `set -xeuo pipefail` comes from though -- there was just a `-e` on the shebang line previously.

From other scripts in this repo, e.g. here:
https://github.com/keycloak/keycloak-containers/blob/master/server/tools/docker-entrypoint.sh#L2

The `-x` there is just to print the expanded form of the statements to be executed. Can be removed if not desired.

keycloak-init-container/extensions.sh
iankko: Ok to keep it uppercase then.

iankko:
@apeschel 

> @iankko Thanks for the review! I've been pretty busy so it took a while to get to following up. I made all your suggested changes, with the exception of the `FAILED_DOWNLOADS` casing, since it's a global variable. Let me know if you would still prefer it to be lowercase.

Thank you for the changes! Ok to keep `FAILED_DOWNLOADS` as uppercase then.

The change LGTM. But could you squash the three commits into one final one, yet?

Thank you.

iankko (changes requested):
LGTM (once the commits are squashed).

apeschel:
@iankko sqauashed

iankko:
> @iankko sqauashed

@apeschel Brilliant, TYVM! One kind request yet though. The `s/KEYCLOAK_EXTENSION/KEYCLOAK_EXTENSIONS/g` issue got addressed within [KEYCLOAK-18143 JIRA](https://issues.redhat.com/browse/KEYCLOAK-18143) already (via https://github.com/keycloak/keycloak-containers/pull/321). 

That means, once you rebase this PR against the current `master`, you will (probably) see a conflict of the PR with the master. Could you rebase to `master` and fix that conflict yet?

Thanks!
Jan

apeschel:
@iankko I already rebased on master when I squashed, there no merge conflict with master :)

iankko (approved):

iankko:
> @iankko I already rebased on master when I squashed, there no merge conflict with master :)

@apeschel Ok, great. Thanks!

drichtarik (approved):

stianst:

PR #320, opened 2021-05-15
Support for ignoring SSL certificate error

kesavkolla:
Added --insecure parameter to ignore SSL certificate errors while downloading the extensions.


iavael:
This doesn't look secure. At least put this behavior behind environment variable flag.

stianst:

PR #322, opened 2021-06-12
Fix errors in postgres example

derekjc:
- The example command for use with postgres is not valid. The docker image should be the last argument in the docker command.
- Without specifying environment variable DB_ADDR, it throws the error `/opt/jboss/tools/docker-entrypoint.sh: line 165: DB_ADDR: unbound variable`
- JIRA: KEYCLOAK-17451


stianst:

PR #324, opened 2021-06-28
Add env variable for importing realm from directory

llka:
I have to use different realm import settings and I don't want to override any XML configurations or oodles of CMD params.
Why don't you add some more useful env variables which will ease keycloak configuration in docker compose?


chris-aeviator:
please merge this!

stianst:

PR #329, opened 2021-10-09
use file_env for DB_ADDR, DB_VENDOR

trigrab:
Hi,

I am using docker secrets for my keycloak deployment. As I am automating everything I also want to set the DB_ADDR via docker secrets. This PR adds the possibility to set DB_ADDR and DB_VENDOR via docker secrets.

I am looking forward to read your comments.


stianst:

PR #330, opened 2021-10-11
Added a default healthcheck to Dockerfile

itay-grudev:
Adds a default `HEALTHCHECK` to be able to monitor the health of the container. A user can always change it should they need to, in their docker-compose file.

I specifically chose a start period of `30s` because the image loads quite slow sometimes, even on SSD.


itay-grudev:
Ping!
@vmuzikar

stianst:

PR #331, opened 2021-10-15
Update README.md

Dannyx323:
Oracle SYSTEM should not be used instead create a new table space with a user and set privledges.

### DIFF
--- a/server/README.md
+++ b/server/README.md
@@ -200,12 +200,31 @@ Alternately, the JDBC file can be copied into the container using the `docker cp
 
 If you used a name for the Oracle instance other than `oracle` you need to specify the `DB_ADDR` environment variable.
 
+#### Prepare Oracle Database
+Never ever use Oracle SYSTEM user!,But create a new table space and add a user with needed priveledges.
+
+Like:
+-- Create a new tablespace KEYCLOAK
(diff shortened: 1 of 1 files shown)


Dannyx323:
> I'd rather just keep it as is and add a small notice that system user shouldn't be used. We shouldn't cover here how you use Oracle, but only how you connect to it. I do somewhat question of this example should be here at all as it is not using an official container from Oracle, but some basic community thing.

Hi stianst,

I would like to point out it is NOT some "basic community thing"  only,  but also applied to a Oracle Exadata database and thats some real database stuff. 
These commands have been checked by a Oracle certified DB.
;-)

stianst:
> > I'd rather just keep it as is and add a small notice that system user shouldn't be used. We shouldn't cover here how you use Oracle, but only how you connect to it. I do somewhat question of this example should be here at all as it is not using an official container from Oracle, but some basic community thing.
> 
> Hi stianst,
> 
> I would like to point out it is NOT some "basic community thing" only, but also applied to a Oracle Exadata database and thats some real database stuff. These commands have been checked by a Oracle certified DB. ;-)

What I question is whether or not the Oracle example should be in here at all as it is using a community container image (https://hub.docker.com/r/carloscastillo/rgt-oracle-xe-11g) not an official image from Oracle. I wasn't not referring to your steps to add the user/etc.

I'm going to close this PR as I said we don't want to document how to use Oracle here. If you'd like to point to some external docs for these steps, or at least add a note that the system user shouldn't be used, that'd be great.

PR #332, opened 2021-10-20
Simple update to JDBC_POSTGRES_VERSION

brx-craig:
Apologies for not sending an email before this PR. Simple version update.

https://issues.redhat.com/browse/KEYCLOAK-19605


brx-craig:
> Please remove the update to KEYCLOAK_VERSION as those updates are handled automatically by our release process. If you could also create a JIRA for the upgrade of Postgres JDBC version that'd be great.

Done and done. https://issues.redhat.com/browse/KEYCLOAK-19605

stianst:

PR #340, opened 2021-11-19
[KEYCLOAK-19831] Include latest Keycloak.X changes in the Docker image

andreaTP:
Co-authored-by: Dominik Guhr <dguhr@redhat.com>


pedroigor:
@DGuhr @andreaTP Awesome work!

@stianst Could you please enable GHA workflow to run in this repo?

pedroigor (approved):

DGuhr (approved):
LGTM ;)

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,25 @@
...
+        uses: actions/checkout@v2
+      -
+        name: Set up QEMU
+        uses: docker/setup-qemu-action@v1
+      -
+        name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v1
+      -
+        name: Build Keycloak.X image
+        uses: docker/build-push-action@v2
```
stianst, line 21: Please pin to commit SHA, as versions can be modified, a review of the actions also has to be done before we add additional community maintained GH actions

andreaTP: @stianst Docker images are now pointing to stable SHAs, regarding the GH Action it's now only using `checkout` at the very same version as the main repo `keycloak/keycloak`.
Do you think this is enough?

DGuhr (approved):

.github/workflows/ci.yaml
andreaTP: @stianst included the specific `sha` even for the checkout action:
https://github.com/keycloak/keycloak-containers/pull/340/files#diff-944291df2c9c06359d37cc8833d182d705c9e8c3108e7cfe132d61a06e9133ddR13

server-x/.gitignore
```diff
@@ -0,0 +1 @@
+*.tar.gz
```
stianst, line 1: Why is this needed?

server-x/Dockerfile
```diff
@@ -1,31 +1,37 @@
-FROM registry.access.redhat.com/ubi8-minimal AS build-env
+FROM registry.access.redhat.com/ubi8-minimal@sha256:16da4d4c5cb289433305050a06834b7328769f8a5257ad5b4a5006465a0379ff AS build-env
```
stianst, line 1: Please remove this change as we don't have a way to handle these updates today, we can also trust these images.

server-x/Dockerfile
```diff
@@ -1,31 +1,37 @@
...
+    tar -xvf /tmp/keycloak/keycloak.x*.tar.gz && \
+    rm /tmp/keycloak/keycloak.x*.tar.gz) || true
+
+RUN mv /tmp/keycloak/keycloak.x* /opt/keycloak
     
-ADD tools /opt/jboss/tools
+COPY tools /opt/keycloak/tools
 
-FROM registry.access.redhat.com/ubi8-minimal
+FROM registry.access.redhat.com/ubi8-minimal@sha256:16da4d4c5cb289433305050a06834b7328769f8a5257ad5b4a5006465a0379ff
```
stianst, line 20: See above

server-x/README.md
```diff
@@ -6,7 +6,7 @@ See the `Build` section for more details on how to build the image.
 
 Once built, you can run the server in the same manner as when using the Keycloak.X distribution by passing any command-line argument:
 
-    docker run --name keycloak -p 8080:8080 \
+    docker run --rm --name keycloak -p 8080:8080 \
```
stianst, line 9: We shouldn't add the "--rm" option here, as someone less observant could copy/paste this and by mistake it would destroy their currently running instance

server-x/tools/docker-entrypoint.sh
```diff
@@ -38,48 +38,6 @@ file_env 'KEYCLOAK_ADMIN_PASSWORD'
...
-    exec /opt/jboss/keycloak/bin/kc.sh config $CONFIG_ARGS &
-    wait $!
-    touch $CONFIG_MARKER_FILE
-  fi
-
-  exec /opt/jboss/keycloak/bin/kc.sh $SERVER_OPTS
-else
-  exec /opt/jboss/keycloak/bin/kc.sh $SERVER_OPTS $CONFIG_ARGS
-fi
+exec /opt/keycloak/bin/kc.sh $@
```
stianst, line 41: Do we even need a "docker-entrypoint.sh" file?

.github/workflows/ci.yaml
stianst: Pinpoint only the GH Actions, not the Docker images (as we don't have a way to keep those updated atm)

stianst:
If we're adding a workflow that can check if the image can be built, would it make sense to go one step further and start it and check one URL is ready?

server-x/README.md
andreaTP: I'm wondering how can this happen?
Named containers needs to be removed before another one can start.

We thought this is an improvement since "someone less observant" will not leave a stopped container dangling.

But not fussy on this detail, following the suggestion to speed up the process 👍

server-x/tools/docker-entrypoint.sh
andreaTP: cc. @pedroigor 
we still have this logic encoded there:
https://github.com/keycloak/keycloak-containers/blob/f8674774f6852110c4e64a30815da770374fa7c9/server-x/tools/docker-entrypoint.sh#L4-L35

server-x/Dockerfile
andreaTP: done

server-x/Dockerfile
andreaTP: done

andreaTP:
> If we're adding a workflow that can check if the image can be built, would it make sense to go one step further and start it and check one URL is ready?

@stianst done a basic check that the endpoint returns 200, do you want to also check the content?
for the rest this PR should be ready for another review 🙂

.github/workflows/ci.yaml
andreaTP: @stianst reverted this change as the only action used here is provided by GH now.

server-x/tools/docker-entrypoint.sh
pedroigor: @stianst It is only there because it seems to be related to a docker feature. To be honest, I don't know how it works but it is there in the wildfly image.

If you think it does not make sense anymore, we can definitely remove the entrypoint script.

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,22 @@
...
+      -
+        name: Build the Keycloak.X image
+        run: |
+          cd server-x
+          docker build . -t keycloakx
+      -
+        name: Test the Keycloak.X image
+        run: |
+          docker run --rm --name keycloak -p 8080:8080 keycloakx start-dev &
+          timeout 300s bash -c 'until curl localhost:8080; do sleep 2; done'
```
okft, line 22: For your information, I'm using the following command to wait on official docker image to be ready:

until curl \
    --head \
    --fail \
    --silent http://127.0.0.1:9990/health > /dev/null
do
  sleep 2
done

(but I know it's WildFly health endpoint, not keycloak one)

server-x/tools/docker-entrypoint.sh
stianst: We have the "file" vault right which allows mounting secrets into files? I'd rather align with that then have these separate docker image only env variables (that won't be documented automatically as we're listing properties from Keycloak source code).

andreaTP:
Updated the PR with the following:
 - add `--fail` to the curl command (thanks for the suggestion @okft !)
 - removed the `docker-entrypoint.sh` cc. @stianst 
 - removed the `probes` folder as unused/untested
 
 I'll squash the commits when I have a nudge that everything is fine.

server-x/tools/docker-entrypoint.sh
pedroigor: Sounds good to me.

stianst (approved):

PR #347, opened 2021-12-16
docker-entrypoint removes add-user.json

ch-lepp:
Docker-entrypoint.sh removes old add-user.json before adding the admin user.
This prevents a bug upon container restart: https://github.com/keycloak/keycloak/issues/9178

### DIFF
--- a/server/tools/docker-entrypoint.sh
+++ b/server/tools/docker-entrypoint.sh
@@ -37,6 +37,12 @@ file_env 'KEYCLOAK_USER'
 file_env 'KEYCLOAK_PASSWORD'
 
 if [[ -n ${KEYCLOAK_USER:-} && -n ${KEYCLOAK_PASSWORD:-} ]]; then
+    
+    add_user_json="/opt/jboss/keycloak/standalone/configuration/keycloak-add-user.json"
+    if [ -e "$add_user_json" ]; then
+        rm "$add_user_json"
+    fi
+    
     /opt/jboss/keycloak/bin/add-user-keycloak.sh --user "$KEYCLOAK_USER" --password "$KEYCLOAK_PASSWORD"
 fi
 


stianst:
Not sure this is the right fix, as it assumes a mounted volume. The container is not supposed to crash if user is already added, which is how it used to work at least. I haven't had time to try it recently though

ch-lepp:
This is not about a crashing container.
This is about stopping (not removing) and restarting a container.

In this case the users get added again (or at least the attempt is made).
However the volume does not get unmounted during any of these steps

jenkincstang:
Hi @stianst 
I also met this issue when redeploy keycloak using docker, would you pls help to check it? thanks

minusdavid:
Also got bitten by this problem. The Keycloak container loses its DB connection due to a routine restart of the DB server, and then it gets stuck in a restart loop.

The problem doesn't happen for a "docker-compose stop; docker-compose start", so the stop must include graceful cleanup that a crash plus restart isn't managing.

minusdavid:
We encounter this because we use a docker-compose restart policy that restarts the container if there's a container crash or a host reboot: https://docs.docker.com/config/containers/start-containers-automatically/

minusdavid:
Also tested using "docker-compose restart" and it worked fine.

Looking at the logs, the hundreds of lines of "User with username 'X' already added to '/opt/jboss/keycloak/standalone/configuration/keycloak-add-user.json'" occurred after a fatal DB error that occurred while Keycloak was trying to shutdown due to a fatal DB connection error during startup.

minusdavid:
It looks like "importAddUser" in KeycloakApplication deletes the addUserFile which is the /opt/jboss/keycloak/standalone/configuration/keycloak-add-user.json file after it tries to add the user:
services/src/main/java/org/keycloak/services/resources/KeycloakApplication.java

However, that deletion won't happen if the Keycloak startup fails.

Keycloak itself doesn't mind if the user already exists:
16:53:01,411 INFO  [org.keycloak.services] (ServerService Thread Pool -- 65) KC-SERVICES0006: Importing users from '/opt/jboss/keycloak/standalone/configuration/keycloak-add-user.json'
16:53:01,792 WARN  [org.keycloak.services] (ServerService Thread Pool -- 65) KC-SERVICES0104: Not creating user X. It already exists.

It looks like the issue is with /opt/jboss/keycloak/bin/add-user-keycloak.sh and https://github.com/keycloak/keycloak-containers/blob/main/server/tools/docker-entrypoint.sh

In add-user-keycloak.sh, we're trying to run createUser but it throws a fatal error if the user is already in the addUserFile
wildfly/adduser/src/main/java/org/keycloak/wildfly/adduser/AddUser.java

Since docker-entrypoint.sh has a "pipefail" setting at the top, the script exits and Keycloak never starts.

minusdavid:
So removing that file in the docker-entrypoint.sh could make sense. 

I'd say the only issue with the PR is that we might not know for sure what the "addUserFile" should be. Maybe it could be something other than "/opt/jboss/keycloak/standalone/configuration/keycloak-add-user.json" although... with the current docker-entrypoint.sh script... I think that would probably cause problems if it were. 

An alternative would be a Java level fix for ignoring the exception, forcing the user into the file, etc.

I suppose other alternatives would be providing one's own custom docker-entrypoint.sh  or mounting one's own  /opt/jboss/keycloak/bin/add-user-keycloak.sh scripts...

minusdavid:
FYI while I wait to hear back on this issue... I copied the docker-entrypoint.sh, added a patch like yours (including a message to say the file was found and removed), and bind mounted it back.

I was able to test it by starting the container, stopping it after the add-user-keycloak.sh ran but before the Keycloak application started up, and it worked a treat. Saw my warning message and everything started up nicely. 

Using the workaround isn't an ideal scenario but it should mitigate the problem for now...

stianst:

PR #348, opened 2021-12-16
CIAM-1173 Upstream 7.5 changes to productize.sh

ASzc:


vmuzikar (approved):
As far as I can tell LGTM.

Issue #365, opened 2022-01-21
Send output of jgroups discovery scripts to stderr/stdout


sandergates:
### Description

If `JGROUPS_DISCOVERY_PROTOCOL` is set, the container (from `jgroups.sh`) will look for a discovery script and run it.  Even if it doesn't exist, a default CLI script will be run.

Currently the output from this script is always sent to /dev/null.  However it seems reasonable people would want to debug the output of the scripts as they run - whether for typos in the script, syntax errors, or configuration/execution problems.

Rather than turn it on for everyone, proposal is to use an environment variable to turn the output on on demand.

### Motivation

Scripts that enable different discovery protocols (e.g. TCPPING) are not trivial or even well documented.  As such, it is normal to iterate on and debug them, but that currently isn't possible if the output is swallowed by the calling jgroups.sh script.  Removing the redirect to /dev/null will allow the output to go to stdout/stderr, and show in the docker logs.

### Details

Proposal is to allow a `JGROUPS_DISCOVERY_CLI_LOGGING` environment variable.  If this is set to `true`, the script output will not be redirected, and will instead go to stdout/stderr as normal.

If the environment variable is not set, or set to anything other than `true`, operation will continue as normal (output will be directed to /dev/null).

stianst:
With Keycloak 20 the WildFly based distribution is no longer supported. For the newer Quarkus distribution of Keycloak, check out the [new documentation](https://www.keycloak.org/server/containers), or the [updated container sources](https://github.com/keycloak/keycloak/tree/main/quarkus/container).

PR #366, opened 2022-01-21
#365 Adding JGROUPS_DISCOVERY_CLI_LOGGING environment variable

sandergates:
If set to true, output from discovery CLI scripts will be sent to stdout/stderr instead of /dev/null.


stianst:

PR #391, opened 2022-04-07
Add support for ppc64le images

mgiessing:
Add support in order to also release ppc64le images. This is no change content-wise.
Thanks!


stianst:

PR #407, opened 2022-06-10
Add -storetype jks to keytool command

archoversight:
On FIPS enabled systems keytool would attempt to use the pkcs11
configuration which is in read only mode, and would fail to create the
trust store.

Bypassing that with the -storetype flag means that Keytool will
successfuly create the keystore.


stianst:

